PR #5, opened 2015-12-16
version number post split

### LINKED ISSUE #3: Version Number Scheme
What version number scheme should ocamlbuild follow?

I created the milestone with the title `4.03` but it can be changed.

@gasche @whitequark

bobot:
Fix #3 by adding `--version` and `--vnum` and using ocamlbuild own 0.9 `VERSION` file

For completing the change of version scheme I will run
```
for i in $(git tag); do
  git tag 0.$i $(git rev-parse $i); 
  git tag -d $i;
  git push -d origin $i; 
  git push origin 0.$i;
done
```

That replace the tag `$i` by `0.$i` on github (It should be done soon if at all since tag shouldn't be moved):
```
git tag 0.4.02.3 7d56b6c14805fe2009dd69e4ca6e0c67ef71e751
git tag -d 4.02.3
git push -d origin 4.02.3
git push origin 0.4.02.3
```


Makefile.create_config
```diff
@@ -24,4 +24,5 @@ src/ocamlbuild_config.ml: force
 	echo 'let so = "$(SO)"'; \
 	echo 'let ext_dll = "$(EXT_DLL)"'; \
 	echo 'let exe = "$(EXE)"'; \
+	echo 'let version = "$(shell cat VERSION)"'; \
```
whitequark: This would break on Windows. I know ocamlbuild doesn't work on Windows anyway, but we could avoid making the situation worse.

whitequark:
Is there any particular reason for `0.9`? The number nine is not special, just use `0.1`...

whitequark:
Ah yeah, I see now.

Makefile.create_config
bobot: We should not, indeed!  But I don't know what is the best way to solve this case and all the others. Depend on external utilities (such as coreutils for windows), rewrite all we need in ocaml (perhaps in a lib installable with ocamlbuild), find specific work around in each case.
Do you have a simple fix for this one?

We should decide in a specific issue in which way we want to solve the windows port. Perhaps there is already one on mantis.

whitequark: This is an one-line OCaml script--I think running it using the ocaml toplevel is perfectly fine.

gasche:
I may have missed this PR when it was submitted. I'm fine with it as long as you both agree on the Windows script stuff. Thanks, François, for giving the commands explicitly -- I'm sure this will turn out to be useful.

You should both have write access to the repository, please do not hesitate to use it when you are confident a merge is a good idea (for example, if two trusted contributors agree on it).

Makefile.create_config
gasche: Any plans to move forward on this? If I understand correctly, this is the only remaining issue for this PR to be merged.

whitequark: `$(ocaml get_version.ml)`
get_version.ml: `print_endline (input_line (open_in "VERSION"))`
?

whitequark: `$(ocaml get_version.ml)`
get_version.ml: `print_endline (input_line (open_in "VERSION"))`
?

bobot: Sorry for the delay. Done.

PR #20, opened 2015-12-23
`ocamlbuild -documentation` support for parametrized flags

gasche:
This is a post-split version of ocaml/ocaml#231. Any feedback is welcome.

Currently the `-documentation` tag only support non-parametrized flag declaration. This has the unfortunate consequence that, for example, the deprecated `warn_A` tag is documented, but not the parametrized `warn(A)` option.

This pull request provides support for documentation of parametrized flags, by providing at declaration time a "documentation instance" for the parameter. The documentation generated corresponds to the rule produced for this particular parameter instance. For example,

    pflag ["ocaml"; "native"; "pack"] "for-pack" ~doc_param:"PackModule"
         (fun param -> S [A "-for-pack"; A param]);

generates the documentation

    parametrized flag {. for-pack(PackModule), native, ocaml, pack .}
      "-for-pack PackModule"

If no documentation parameter is provided, we pass the parameter `X` by default, and catch any exception raised.


rgrinberg:
Did the 2nd last commit sneak in here somehow? I'm not quite sure how it's related.

I could be just cargo culting here, but I thought the preference for optional arguments was to put them ahead of any non optional arguments.

src/flags.mli
```diff
@@ -28,7 +28,27 @@ val of_tag_list : Tags.elt list -> Command.spec
    has a real need for deprecation, drop us a note on the bugtracker. *)
 val flag : ?deprecated:bool -> Tags.elt list -> Command.spec -> unit
 
-val pflag : Tags.elt list -> string -> (string -> Command.spec) -> unit
+(* The ?doc_param parameter provides an example parameter that will be
+   used to generate documentation for this deprecated flag: we invoke
```
bobot: `deprecated` -> `parameterized`

src/flags.mli
```diff
@@ -28,7 +28,27 @@ val of_tag_list : Tags.elt list -> Command.spec
...
+   at least fail to parse a dummy parameter), but just list its
+   activation tags in the documentation -- so users will still be
+   aware of its existence.
+
+   As for the ?deprecated flag above, this is not exported to the
+   plugin interface for now. *)
+val pflag : Tags.elt list -> string ->
+            ?doc_param:string -> (string -> Command.spec) -> unit
+
+val flag_and_dep : Tags.elt list -> Command.spec -> unit
```
bobot: doc?
```
(** same as {!flag} except that pathnames that appear in the command spec are automatically added as
dependency when this flag is added *)
```

src/flags.mli
```diff
@@ -28,7 +28,27 @@ val of_tag_list : Tags.elt list -> Command.spec
...
+
+val pflag_and_dep : Tags.elt list -> Tags.elt ->
+  ?doc_param:string -> (string -> Command.spec) -> unit
```
bobot: doc?

src/ocaml_specific.ml
```diff
@@ -296,7 +298,15 @@ rule "ocaml C stubs: clib & (o|obj)* -> (a|lib) & (so|dll)"
...
-  ?doc:None (* TODO document *)
+  ~doc:"OCamlbuild can create a C library by calling ocamlmklib \
+        with the file paths listed (one per line) in libfoo.clib. \
+        To build a static library from libfoo.clib, you should \
+        request libfoo.a (or libfoo.lib on Windows), and to build \
+        a dynamic library you should request libfoo.so (or libfoo.dll \
+        on Windows). Finally, any file listed in the .clib \
+        with name 'bar/baz.o' will link 'bar/baz.obj' instead on Windows: \
+        using the .o extension will give portable clib files, even if \
+        it is not the object file extension on your system."
```
bobot: Nice to fill these todo. Just `instead on Windows: using the` -> `instead on Windows. Indeed using the ...` for simplicity.

src/ocaml_specific.ml
```diff
@@ -430,14 +440,20 @@ rule "ocamldoc: document ocaml project odocl & *odoc -> docdir (man)"
   ~prod:"%.docdir/man"
   ~stamp:"%.docdir/man.stamp"
   ~dep:"%.odocl"
-  ?doc:None (* TODO document *)
+  ~doc:"If foo.odocl contains a list of capitalized module names, \
+        the target foo.docdir/man will call ocamldoc to produce \
+        documentation for this modules in man format."
```
bobot: I prefer the positive form that you use for `.clib`.

```
~doc:"The target foo.docdir/man will call ocamldoc to produce \
          documentation in man format for the modules listed in \
          capitalized form in foo.odocl."
```

src/ocaml_specific.ml
```diff
@@ -430,14 +440,20 @@ rule "ocamldoc: document ocaml project odocl & *odoc -> docdir (man)"
...
-rule "ocamldoc: document ocaml project odocl & *odoc -> man|latex|dot..."
+rule "ocamldoc: document ocaml project odocl & *odoc -> latex|dot..."
   ~prod:"%(dir).docdir/%(file)"
   ~dep:"%(dir).odocl"
-  ?doc:None (* TODO document *)
+  ~doc:"If foo.odocl contains a list of capitalized module names, \
+        the target foo.docdir/bar.tex (or bar.ltx) will call ocamldoc \
+        to produce documentation for these modules in LaTeX format. \
+        The target foo.docdir/bar.dot will generate a dot graph of module \
+        dependencies."
```
bobot: idem for positive form.

src/ocaml_specific.ml
```diff
@@ -489,7 +505,11 @@ end else
 rule "ocaml C stubs: c -> o"
   ~prod:x_o
   ~dep:"%.c"
-  ?doc:None (* TODO document *)
+  ~doc:"The OCaml compiler can be passed .c files and will call \
+        the underlying C toolchain to produce corresponding .o files. \
+        Requiring foo.o from OCamlbuild may build it from foo.c. \
```
bobot: `Requiring foo.o from OCamlbuild may build it from foo.c` -> `This rule gives the possibility to use this transformation.`

gasche:
Thanks to both for your feedback!

> Did the 2nd last commit sneak in here somehow? I'm not quite sure how it's related.

Indeed, I will commit it separately.

> I could be just cargo culting here, but I thought the preference for optional arguments was to put them ahead of any non optional arguments.

Optional arguments must be before at least one mandatory argument, but otherwise the order does not matter. In particular `pflag ~doc_param:None` can be used to pre-apply the argument and obtain a value with the exact same type as the previous `pflag` function, which can be useful in some higher-order use cases.

I put the parameter in this order because I think this is the most natural at the call site. I prefer to read

```ocaml
 pflag ["ocaml"; "native"; "compile"] "for-pack" ~doc_param:"PackModule"
   (fun param -> S [A "-for-pack"; A param]);
```

rather than

```ocaml
 pflag ~doc_param:"PackModule" ["ocaml"; "native"; "compile"] "for-pack"
   (fun param -> S [A "-for-pack"; A param]);
```

(More precisely, it is nice to have it right after the name of the parametrized rule, as `"for-pack" ~doc_param:"PackModule"` suggests the applied form `for-pack(PackModule)`.)

I'll fix @bobot's inline comments.

> I'm not sure it is always easy to see what is parameterized.

One way to solve this issue which would maybe be lighter (also I don't like `e.g.` which is a bit pedantic) would be to always put the parametrized tag first in the list. This requires some different plumbing in the implementation, but it should be easy to do.

src/ocaml_specific.ml
```diff
@@ -560,7 +582,7 @@ flag [ "ocaml" ; "menhir" ; "infer" ] (S[A "--infer"]);;
 let () =
   List.iter begin fun mode ->
     flag [ mode; "only_tokens" ] (S[A "--only-tokens"]);
-    pflag [ mode ] "external_tokens" (fun name ->
+    pflag [ mode ] ~doc_param:"TokenModule" "external_tokens" (fun name ->
```
gasche: reorder param order for consistency

src/flags.mli
gasche: This is already documented in `signatures.mli`.

src/ocaml_specific.ml
gasche: I just removed this confusing line.

gasche:
I committed the `-man` and `-dot` options to master directly, and rebased the PR to take @bobot 's comments into account. The last one remaining is to try to show the parametrized name first in the tag output.

gasche:
The last commit makes the choice of always sorting applied parametrized tags first in `Tags` sets. The documentation output is now as follows:

```
parametrized flag {. man_section(X), doc, manpage, ocaml .} "-man-section X"

parametrized flag {. runtime_variant(_pic), link, ocaml .}
  "-runtime-variant _pic"

parametrized flag {. open(MyPervasives), compile, ocaml .}
  "-open MyPervasives"

parametrized flag {. warn_error(+10+40), compile, ocaml .}
  "-warn-error +10+40"

parametrized flag {. warn(A@10-28@40-42-45), compile, ocaml .}
  "-w A@10-28@40-42-45"

parametrized flag {. ppx(my_ppx), infer_interface, ocaml .} "-ppx my_ppx"
parametrized flag {. pp(my_pp), infer_interface, ocaml .} "-pp my_pp"

parametrized flag {. color(X), compile, ocaml .} "-color X"

parametrized flag {. inline(5), compile, native, ocaml .} "-inline 5"
```

@bobot , does that solve your readability issue with the previous presentation?

PR #24, opened 2015-12-23
Added a Changes file.

Armael:
This is for the moment only the concatenation of OCaml's Change file, and
ocamlbuild's old ChangeLog.


gasche:
I merged your PR and filtered the OCaml changelog. The tools of the trade is `C-u M-|` in Emacs, with the commands

    grep ocamlbuild # for older Changes section with one change per line
    grep ocamlbuild -A 1 --no-group-separator # for more recent Changes section with credit lines

PR #25, opened 2015-12-23
improve documentation of Echo

### LINKED ISSUE #19: clarify argument to Echo
`Echo` prints a string list to a given file. Are the strings intended to be *lines*? If so, I'd expect a newline character to be added automatically, but it isn't. On the other hand, if they aren't meant to be lines, then I wonder why the argument is a `string list` instead of just `string`.

agarwal:
Closes #19.


gasche:
Fine, thanks!

Issue #28, opened 2015-12-24
Missing findlib META file


rgrinberg:
This is required for a proper installation with opam

bobot:
It is better, but I don't see why it is required.

rgrinberg:
Agreed. It's technically optional. It is required for the ocamlbuild library used by plugins however.

gasche:
Fixed by #41.

PR #29, opened 2015-12-24
Local opam files for ocamlbuild

rgrinberg:
Not ready yet because there's no META file

Needs #28 to proceed I think?

Also, does it make sense to add findlib as a dependency in this case as well? I guess it's an optional dependency?

### DIFF
--- a/META
+++ b/META
@@ -1,6 +1,6 @@
 # Specification for the "ocamlbuild" library
 requires = "unix"
-version = "[distributed with Ocaml]"
+version = "0.9"
 description = "ocamlbuild support library"
 directory= "^"
 archive(byte) = "ocamlbuildlib.cma"
--- a/META
+++ b/META
@@ -1,0 +1,8 @@
+# Specification for the "ocamlbuild" library
+requires = "unix"
+version = "[distributed with Ocaml]"
+description = "ocamlbuild support library"
+directory= "^"
+archive(byte) = "ocamlbuildlib.cma"
+archive(native) = "ocamlbuildlib.cmxa"
+
--- a/ocamlbuild.install
+++ b/ocamlbuild.install
@@ -1,0 +1,4 @@
+bin: [
+  "?ocamlbuild.native" {"ocamlbuild"}
+  "ocamlbuild.byte" {"ocamlbuild.byte"}
+]
--- a/opam
+++ b/opam
@@ -1,0 +1,28 @@
+opam-version: "1.2"
+name: "ocamlbuild"
+maintainer: "Gabriel Scherer <gabriel.scherer@gmail.com>"
+
+authors: [
+  "Nicolas Pouillard"
+  "Berke Durak"
+]
+
+license: "LGPL-2 with OCaml linking exception"
+dev-repo: "https://github.com/ocaml/ocamlbuild.git"
+homepage: "https://github.com/ocaml/ocamlbuild/"
+bug-reports: "https://github.com/ocaml/ocamlbuild/issues"
+
+build: [
+  [make]
+  [make "ocamlbuild.native"]
+]
+
+# TODO
+# install: [ ]
+
+doc: [
+  "http://caml.inria.fr/pub/docs/manual-ocaml/ocamlbuild.html"
+  "https://github.com/gasche/manual-ocamlbuild/blob/master/manual.md"
+]
+
+available: [ocaml-version >= "4.03"]
--- a/opam
+++ b/opam
@@ -17,8 +17,8 @@ build: [
   [make "ocamlbuild.native"]
 ]
 
-# TODO
-# install: [ ]
+install: [make "findlib-install"]
+remove: ["ocamlfind" "remove" "ocamlbuild"]
 
 doc: [
   "http://caml.inria.fr/pub/docs/manual-ocaml/ocamlbuild.html"
@@ -26,3 +26,5 @@ doc: [
 ]
 
 available: [ocaml-version >= "4.03"]
+
+depends: [ "ocamlfind" {build} ]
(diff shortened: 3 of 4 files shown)


opam
```diff
@@ -0,0 +1,28 @@
...
+
+# TODO
+# install: [ ]
+
+doc: [
+  "http://caml.inria.fr/pub/docs/manual-ocaml/ocamlbuild.html"
+  "https://github.com/gasche/manual-ocamlbuild/blob/master/manual.md"
+]
+
+available: [ocaml-version >= "3.10"]
```
bobot: I'm not sure this version compile with ocaml 3.10. So in one case it is arbitrary, in the other it is too permissive. So we should remove it or fix something reasonable.

rgrinberg: I took that constraint from the opam file in ocaml/opam-repository. I haven't tested it myself.

bobot: This one is for the built-in ocamlbuild that appeared in 3.10. We should add `< 4.03` in it. And `>= 4.03` in this one.

rgrinberg: OK that makes sense. It's a bit unsatisfactory since I'm sure ocamlbuild can be compiled with ocaml older than 4.02.3 but installing it might interfere with what's already provided.

There's also the option of specifying a conflict with OCaml `< 4.03` but I don't know how that's any better.

gasche:
In a perfect world I would like `ocamlbuild` to *not* depend on `ocamlfind`, because we have had reports from users that insist on this ability.

In the short term I think that making this dependency optional makes packaging too difficult. We should aim for proper packaging for 99% of our users as soon as possible, and refine the dependency structure later -- with inputs from no-ocamlfind users if we can contact them.

Could you add a META file to this pull request?

We could have an `install` rule that just call `make install`, or we could have an explicit `ocamlfind` call in the opam package description. I have no strong opinion on this.

rgrinberg:
> We could have an install rule that just call make install, or we could have an explicit ocamlfind call in the opam package description. I have no strong opinion on this.

We can have explicit install target for ocamlbuild but it should only do findlib installation. Binaries should be left for opam to install (using ocamlbuild.install).

I think this is what you're saying, so I'm just confirming.

avsm:
> In the short term I think that making this dependency optional makes packaging too difficult.

I'm confused -- why does this make packaging a separate `ocamlbuild` more difficult?  When it was integrated with the compiler, `ocamlfind` was clearly optional, so it should remain so now right?  All the OPAM packages that depend on `ocamlfind` should have it marked as an explicit dependency already.

gasche:
@avsm My suggestion was to shoot for a simple packaging first, using `ocamlfind install` to install the ocamlbuild library, and `ocamlbuild.install` to install the opam binary.

I don't know how to make an opam package with an optional ocamlfind dependency (how should we test whether ocamlfind is available?). If you or @rgrinberg or someone else knows, and wants to submit a pull request to put it in, it is of course welcome.

Another question is whether we can have an install target in the `Makefile` that works for people that do *not* use OPAM. It would be nice to have.

avsm:
> Another question is whether we can have an install target in the Makefile that works for people that do not use OPAM. It would be nice to have.

This would be essential for distro packaging, or else they'll all have to replicate the installation logic.

avsm:
> My suggestion was to shoot for a simple packaging first, using ocamlfind install to install the ocamlbuild library, and ocamlbuild.install to install the opam binary.

If I remember right, cmdliner doesn't depend on ocamlfind any more and just installs the META file manually.  `ocamlfind install` doesnt do much in the simple case beyond that.

dbuenzli:
@avsm is correct, all my packages that do not have any dependency do not depend on `ocamlfind` and I never use `ocamlfind install` anyway: I always install the `META` file in `lib` using a `.install` generated by `topkg`.

gasche:
Very good, we should use this technique in our packaging as well.

@rgrinberg would you be motivated to come up with a META file and a Bünzli-inspired `.install` file?

rgrinberg:
I just pushed out a findlib based solution which I've done before the discussion above. Will redo to make ocamlfind optional. Just to re-iterate, we'll use topkg right?

@gasche Check out the META meanwhile.

META
```diff
@@ -0,0 +1,8 @@
+# Specification for the "ocamlbuild" library
+requires = "unix"
+version = "[distributed with Ocaml]"
```
gasche: I think we should rather use our own version numbers, the current choice seems to be 0.9, according to #3 and #5 .

ocamlbuild.install
```diff
@@ -0,0 +1,4 @@
+bin: [
+  "?ocamlbuild.native" {"ocamlbuild"}
+  "ocamlbuild.byte" {"ocamlbuild.byte"}
+]
```
gasche: If I understand correctly the [.install documentation](https://opam.ocaml.org/doc/Manual.html#packagenameinstall), the semantics of this description is that if ocamlbuild.native exists, then it is installed as "ocamlbuild", and ocamlbuild.byte is always installed as ocamlbuild.byte.

This is a bit disappointing: we would like users with no native compiler available to have an `ocamlbuild` executable available, named `ocamlbuild.byte`. I guess that we will need some sort of configuration layer producing the `ocamlbuild.install` file based on the detected native support.

rgrinberg: Agreed, that is how I understood it as well. Perhaps it's worth making an opam issue as well?

gasche:
> Just to re-iterate, we'll use topkg right?

I'm not sure what is best. If we could just use a `.install` file listing what we want to install, that would be the simplest solution (the list of files installed by ocamlbuild is simple and well-defined). If you are familiar with topkg or willing to use it, we can also try that.

I would be ready to merge the first version of this PR that feels ready, even if it still relies on ocamlfind: we can refine the packaging as a second step, but having something that works for our user is a good first milestone ( especially in the light of #32 ). Of course if you prefer to do both in one fell swoop, that is also fine.

gasche:
This PR was integrated by #41 which was merged. Thanks!

PR #30, opened 2015-12-25
[RFC] warn on second call to the dispatch function

gasche:
It is an embarrassing shortcoming of the plugin-tags work that calling the `dispatch` function several times just silently overrides its behaviour. Forcing users to call `dispatch` exactly once may be the right interface to expose, but not saying anything if they make a mistake is bad usability.

One approach would be to keep a queue of hooks, push each new hook, and call them in order at dispatch time. I decided _not_ to do that, however, because that would encourage an imperative way to provide functionality, with each plugin doing its own `dispatch` call; this would be fairly fragile as (1) users would have little control over the ordering on plugin rules and (2) they would have to jump through hoops to make sure the plugins are really linked, otherwise their effect would get silently dropped.

By forcing that there is at most one dispatch call, we in fact enforce a pure, declarative interface between plugin providers and plugin users. This has worked out rather well in practice (despite the subpar usability of the current code), see [js_of_ocaml ocamlbuild plugin](https://ocsigen.org/js_of_ocaml/api/Ocamlbuild_js_of_ocaml) for example.

This PR keeps this current interface of having a single dispatch call, but adds a warning each time a previously installed hook is discarded. (The formatted warning text is rather painful to read, so I will probably convert it to one of Daniel's Bünzli nice formatting functions instead.)

### DIFF
--- a/src/hooks.ml
+++ b/src/hooks.ml
@@ -21,8 +21,26 @@ type message =
   | Before_rules
   | After_rules
 
-let hooks = ref ignore
+let hooks = ref None
 
-let setup_hooks f = hooks := f
+let setup_hooks f =
+  match !hooks with
+  | None -> hooks := Some f
+  | Some _ ->
+      Log.eprintf "%a" Format.pp_print_text
+        "Warning: your myocamlbuild.ml plugin seems to be setting \
+         up several dispatch hooks through several calls to \
+         Ocamlbuild_plugin.dispatch. This is not supported, \
+         all dispatch functions but the last one will be discarded.\n\
+         \n\
+         You should not install several hook handlers, but rather \
+         combine several handler functions explicitly in a single \
+         dispatch call. This lets you declare the order between \
+         sub-hooks instead of relying on some implicit evaluation \
+         effect order.\n";
+      hooks := Some f
 
-let call_hook m = !hooks m
+let call_hook m =
+  match !hooks with
+  | None -> ()
+  | Some hook -> hook m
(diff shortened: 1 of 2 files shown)


src/hooks.ml
```diff
@@ -21,8 +21,23 @@ type message =
...
+  match !hooks with
+  | None -> hooks := Some f
+  | Some _ ->
+      Log.eprintf "Warning:@ you@ are@ setting@ up@ a@ dispatch@ hook@ \
+                   while@ one@ has@ already@ been@ setup;@ the@ previous@ hook@ \
+                   will@ be@ discarded.@\n\
+                   One@ should@ not@ setup@ hooks@ several@ times,@ but@ rather@ \
+                   combine@ several@ dispatch@ functions@ explicitly@ \
+                   in@ a@ single@ dispatch@ call.@ This@ lets@ you@ declare@ \
+                   the@ horder@ between@ sub-hooks@ instead@ of@ relying@ \
```
whitequark: declare the **b**order

whitequark:
@gasche I'd like to nominate this for inclusion in 0.11.

gasche:
I rebased, used `Format.pp_print_text` (available since OCaml 4.02, we require 4.03 and above), and updated the warning message. Would you have any comment on the message? I would like to be not too confusing for beginners, and in particular indicate what has to be fixed.

whitequark:
@gasche can you rebase this on top of master, to trigger the build?

hhugo:
Should we merge this one ?

kit-ty-kate:
Sounds reasonable. Let's

kit-ty-kate:
Thanks!

Issue #14, opened 2015-12-22
improve handling of symlinks in repo root


agarwal:
I've never been happy with how symlinks are created in the repo root. I think this should be off by default. Users should be in control of which files are created, even when those files are symlinks. In particular it is irritating to have `foo.byte` and `foo.native` links created. I certainly don't want both, and I actually I don't want either one.

On the other hand, symlinks I do want get deleted!
```
$ cat myocamlbuild.ml 
open Ocamlbuild_plugin

let () = dispatch (function
  | After_rules ->
     rule ".merlin"
       ~prod:".merlin"
       (fun _ _ -> Echo (["PKG core"],".merlin"))
  | _ -> ()
)

$ ocamlbuild .merlin

$ ln -s _build/.merlin

$ ls -l .merlin
lrwx------  1 ashish  staff    14B Dec 22 15:13 .merlin -> _build/.merlin

$ ocamlbuild -clean

$ ls -l .merlin
ls: .merlin: No such file or directory
```

gasche:
The current semantics, which I don't think has ever changed since ocamlbuild was released, is to create links only on some specific targets (file extensions): `byte`, `native`, `top` (toplevels), and `html` targets (it then creates a link to the `.docdir` directory rather than html file itself). See [the implementation in main.ml](https://github.com/ocaml/ocamlbuild/blob/master/src/main.ml#L256-L260).

Note that you can use the `-no-links` option to never create links. We do *not* have a `-make-links` option (enabled by default), which mean that right now we cannot disable links by default as our users would have not explicit solution to re-enable them. We should add it as soon as possible.

The question of user-made links into `_build` is more subtle. OCamlbuild consider that anything in `_build` is its own territory, and in particular links will always become stale after `ocamlbuild -clean`, so arguably removing them at cleaning time is a correct move. I think we need to discuss your need of generating `.install` and `.merlin` file on its own: I think this is an important and useful feature to have, but we may need a broader design discussion than just changing some aspects of ocamlbuild one at a time to create one of the possible implementation paths.

agarwal:
> Note that you can use the `-no-links` option to never create links.

[...]

PR #35, opened 2015-12-27
Use tags that do not directly depend on the target extension for ocamldoc rules

Armael:
- the generic `%(file)` rule has been split into target-specific ones;
- unhardcode a few things: a symlink to the produced documentation directory was only created when building a "*.html" target. I chose a conservative approach, and it is now created for any ocamldoc rule that builds a directory (atm, html and manpages). This could be easily changed to also create a symlink when producing `.tex`/`.dot` files.

*This PR breaks existing code*. The case I know is when using a custom ocamldoc plugin (eg [ocaml-libudev](https://github.com/Armael/ocaml-libudev/blob/master/myocamlbuild.ml#L40), [tyxml](https://github.com/ocsigen/tyxml/blob/master/myocamlbuild.ml#L57)); the fix would be to replace `"extension:html"` with `"htmldoc"`.

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -430,33 +430,55 @@ rule "ocamldoc: document ocaml project odocl & *odoc -> docdir (html)"
   ~stamp:"%.docdir/html.stamp"
   ~dep:"%.odocl"
   ~doc:"the target foo.docdir/index.html will call ocamldoc to produce \
-        the html documentation for the module paths listed
+        the html documentation for the module paths listed \
         in foo.odocl, one per line. \
-        See also the max|latex|doc target below."
+        See also the man|latex|texinfo|dot target below."
   (Ocaml_tools.document_ocaml_project
+      ~tags:["html"]
       ~ocamldoc:Ocaml_tools.ocamldoc_l_dir "%.odocl" "%.docdir/index.html" "%.docdir");;
 
 rule "ocamldoc: document ocaml project odocl & *odoc -> docdir (man)"
   ~prod:"%.docdir/man"
   ~stamp:"%.docdir/man.stamp"
   ~dep:"%.odocl"
-  ~doc:"The target foo.docdir/man will call ocamldoc to produce
-        documentation, in man format, for the module paths
+  ~doc:"The target foo.docdir/man will call ocamldoc to produce \
+        documentation, in man format, for the module paths \
         listed in foo.odocl, one per line."
   (Ocaml_tools.document_ocaml_project
-      ~tags:["manpage"]
+      ~tags:["man"]
       ~ocamldoc:Ocaml_tools.ocamldoc_l_dir "%.odocl" "%.docdir/man" "%.docdir");;
 
-rule "ocamldoc: document ocaml project odocl & *odoc -> latex|dot..."
-  ~prod:"%(dir).docdir/%(file)"
+rule "ocamldoc: document ocaml project odocl & *odoc -> latex"
+  ~prod:"%(dir).docdir/%(file).%(ext:\"tex\" or \"ltx\")"
   ~dep:"%(dir).odocl"
-  ~doc:"The target foo.docdir/bar.tex (or bar.ltx) will call ocamldoc
-        to produce documentation, in LaTeX format, for the module paths
-        listed in foo.odocl, one per line. \
-        The target foo.docdir/bar.dot will generate a dot graph of module \
+  ~doc:"The target foo.docdir/bar.tex (or bar.ltx) will call ocamldoc \
+        to produce documentation, in LaTeX format, for the module paths \
+        listed in foo.odocl, one per line."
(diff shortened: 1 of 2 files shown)


src/ocaml_specific.ml
```diff
@@ -456,7 +457,36 @@ rule "ocamldoc: document ocaml project odocl & *odoc -> latex|dot..."
...
         dependencies."
   (Ocaml_tools.document_ocaml_project
-      ~ocamldoc:Ocaml_tools.ocamldoc_l_file "%(dir).odocl" "%(dir).docdir/%(file)" "%(dir).docdir");;
+      ~tags:["texdoc"]
+      ~ocamldoc:Ocaml_tools.ocamldoc_l_file
+        "%(dir).odocl" "%(dir).docdir/%(file).tex" "%(dir).docdir");;
+
+rule "ocamldoc: document ocaml project odocl & *odoc -> latex (.ltx)"
+  ~prod:"%(dir).docdir/%(file).ltx"
+  ~dep:"%(dir).odocl"
```
gasche: wouldn't it be possible to factorize the `.tex` and `.ltx` rules? From the top of my head I don't remember the kind of glob patterns we have available.

src/ocaml_specific.ml
```diff
@@ -456,7 +457,36 @@ rule "ocamldoc: document ocaml project odocl & *odoc -> latex|dot..."
...
+  ?doc:None (* TODO document *)
```
gasche: I'm not sure whether this is a remnant from a patch developed before I merged #20, or whether it is because you added new rules, but in any case you must fix this :-)

src/ocaml_tools.ml
```diff
@@ -120,7 +120,10 @@ let ocamldoc_l_dir tags deps _docout docdir =
       Cmd (S[A"mkdir"; A"-p"; Px docdir]);
       Cmd (S [!Options.ocamldoc;
               S(List.map (fun a -> S[A"-load"; P a]) deps);
-              T(tags++"doc"++"docdir"); A"-d"; Px docdir])]
+              T(tags++"doc"++"docdir"); A"-d"; Px docdir]);
+      Cmd (S [A "ln"; A "-sf";
+              P (Pathname.concat !Options.build_dir docdir);
+              A Pathname.pwd])]
```
gasche: I'm not sure I like the fact of spreading out the symbolic-linking logic in more places (there is already some in this file to handle `.itarget` rule). This aspect of ocamlbuild is currently doubtful (see #14 ). In particular, your implementation does not respect the `-no-links` option, so it is a regression on this front.

For now I think the simplest step forward would be to drop this part of the PR, and concentrate on good documentation targets.

gasche:
Do I understand correctly that, if the `ocalmdoc` issue that custom generation and `-html` do not mix well is fixed upstream, then this PR will not break existing projects anymore? Breaking projects is very bad and I'm not sure the gains are worth it, so if this understanding is correct I'll wait until the ocamldoc side is fixed to merge.

Do we get to have a nice bike-shedding session about the tag name you chose? Currently you use `htmldoc`, `manpage`, `depgraph`, `texdoc` and `texidoc`. That does not seem particularly consistent. Maybe we could just use the format (`html`, `manpage` (we could also support `man` and keep `manpage` only for backward compatibility), `dot`, `tex`, `texi`)?

Note that I merged 5b022ddc569e0d1f65a5efccc234c5809f82786e which adds `manpage`- and `extension:dot`-specific tags, and may inform the naming choice. By the way, it's strange that you did not replace `extension:dot` by `depgraph` in these as well.

Armael:
I fixed the inline comments stuff.

>Do I understand correctly that, if the ocamldoc issue that custom generation and -html do not mix well is fixed upstream, then this PR will not break existing projects anymore? 

Correct.

>Breaking projects is very bad and I'm not sure the gains are worth it, so if this understanding is correct I'll wait until the ocamldoc side is fixed to merge.

Ok.

>Do we get to have a nice bike-shedding session about the tag name you chose? Currently you use htmldoc, manpage, depgraph, texdoc and texidoc. That does not seem particularly consistent. Maybe we could just use the format (html, manpage (we could also support man and keep manpage only for backward compatibility), dot, tex, texi)?

We do not get to bike-shed. I'm not really good with names, so I happily changed to what you suggest.

>By the way, it's strange that you did not replace extension:dot by depgraph in these as well.

I think I did?

gasche:
Very good! I'm waiting for you to move on ocamldoc upstream :-)

gasche:
P.S.: no need to use #N for issues and PR#N for pull requests, you should use #N consistently. I'll maybe change this mention in the Changelog that seems confusing.

Armael:
Oh, I didn't know issues and PRs numbers were in the same scope. I fixed the Changes entry.

Also, mantis ticket created: http://caml.inria.fr/mantis/view.php?id=7108 .

PR #37, opened 2015-12-27
generic dependencies: 'link_dep' linking dependency rule

gasche:
OCamlbuild comes with a fair number of built-in flags and parametrized
flags (flag, pflag), but with no built-in dependency rule. This
omission makes it difficult in particular for people to understand how
to build mixed OCaml/C program, as they have to add their own linking
rules (for C object modules) to myocamlbuild.ml.

This commit goes a first step towards more usable generic dependency
rules by adding a generic link-time dependency. In _tags,

    pred: link_dep(foo)

will add "foo" as dependency to all targets matching "pred", and add
"foo" as a link-time argument to linking actions performed to produce
those targets. In particular,

    "foo.byte": link_dep(foo.o), custom
    "foo.native": link_dep(foo.o)

is what should be used to link C stubs into an OCaml executable.

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -789,7 +789,6 @@ flag ["ocaml"; "absname"; "infer_interface"] (A "-absname");;
 flag ["ocaml"; "byte"; "compile"; "compat_32";] (A "-compat-32");;
 flag ["ocaml";"compile";"native";"asm"] & S [A "-S"];;
 
-
 (* threads, with or without findlib *)
 flag ["ocaml"; "compile"; "thread"] (A "-thread");;
 flag ["ocaml"; "link"; "thread"] (A "-thread");;
@@ -851,5 +850,8 @@ pflag ["ocaml"; "doc"; "manpage"] "man_section"
 
 ocaml_lib "ocamlbuildlib";;
 ocaml_lib "ocamlbuildlightlib";;
+
+(* generic dependency rules *)
+pdep ["link"; "link_with"] "link_dep" (fun param -> [param]);;
 
 end in ()
--- a/testsuite/internal.ml
+++ b/testsuite/internal.ml
@@ -321,4 +321,25 @@ let () = test "OpenDependencies"
   ~matching:[M.f "b.byte"]
   ~targets:("b.byte",[]) ();;
 
+let () = test "LinkCStubs"
+  ~description:"Build a simple project that links with C stubs"
+  ~options:[`no_ocamlfind]
+  ~tree:[
+    T.f "a.ml" ~content: "external hello_world : unit -> unit = \"hello_world\";; let () = hello_world ();;";
+    T.f "b.c" ~content:{|
+#include <stdio.h>
+#include <caml/mlvalues.h>
+#include <caml/memory.h>
+CAMLprim value hello_world(value unit)
+{
+  CAMLparam1 (unit);
+  printf("Hello World!\n");
+  CAMLreturn (Val_unit);
+}
+|};
+    T.f "_tags" ~content:{|"a.byte": link_dep(b.o),custom|};
+  ]
+  ~matching:[M.f "a.byte"]
+  ~targets:("a.byte", []) ();;
+
 run ~root:"_test_internal";;


gasche:
The idea for this came from observing the requests in gasche/manual-ocamlbuild#14 to document how to build mixed OCaml/C modules: the best way to document something is by first making it as easy as possible.

gasche:
I would be interested by some feedback from people that actually have experience building mixed OCaml/C program with OCamlbuild. Do you use such a generic rule, and maybe have a recommendation on the naming? @bobot, @whitequark, @yallop ?

yallop:
I don't use ocamlbuild, but @dsheets uses it to build mixed OCaml/C programs and might have suggestions.

whitequark:
Good idea; I have a variation of this in nearly every project that uses C stubs. Would be best if there would be another flag (or maybe even the same?) that adds a `-lfoo` to the flags embedded in a `.cma`.

gasche:
@whitequark: I don't know the details of distributing `.cma` containing modules that depend on C stubs (the `.o` files have to be distributed along with the `.cma`, I suppose?). It would be useful if you could craft a small example of `.cma` production that does the right thing, as a testsuite that fails if `-lfoo` is not passed properly or `foo.o` not put at the right place.

whitequark:
No no, no one distributes or links in `.o` files. The way you do it is you put `foo.o` in `libfoo.clib`, which creates a target `libfoo.a`, which you then distribute, and which is autolinked by `-lfoo` in the `.cma`.

gasche:
Thanks, I could reproduce this workflow at home.

I would find it odd for a parametrized rule with parameter `foo.o` to strip extensions and add a `-lfoo` tags, so maybe the interface we are looking for is `link_c_lib(foo)` that builds and link `libfoo.a` to binaries and compiles archives with `-lfoo`. This would be the "more advanced but ultimately more convenient way", because it requires writing a `libfoo.clib`, and `link_dep(foo.o)` would be the "fast and quick" way. I can document both.

I'll try to write such a `link_clib` rule and update the PR accordingly.

whitequark:
I agree on `link_c_lib(foo)`, but I am not sure what is the usefulness of bare `link_dep`. In case of a bytecode library, what makes the invocation of the compiler that consumes the library link in `foo.o`?

gasche:
> I am not sure what is the usefulness of bare `link_dep`.

I tend to agree. Furthermore, we could add a rule `%.c -> lib%.a` that corresponds to a single-module `.clib` file, and then the `link_c_lib(foo)` method would subsume `link_dep(foo.o)`.

> In case of a bytecode library, what makes the invocation of the compiler that consumes the library link in foo.o?

I'm not sure I understand the question. I understand it as: what should appear in the source program to force linking? with `.cma` archives, the module has to be depended upon explicitly unless `-linkall` is passed. My answer to this question is: I don't know -- I would guess that they are always linked.

whitequark:
> Furthermore, we could add a rule %.c -> lib%.a

That would be great.

> I'm not sure I understand the question.

I mean--if I pass `-lfoo` while creating a `.cma`, autolink picks it up and it's enough to depend on the OCaml library. If I want consumers of my library to link in a `foo.o`, they have to do so explicitly, since autolink only works on library level as far as I'm aware.

gasche:
Yeah; with the `link_c_lib` scheme we would encourage people to deal with `.a`, which sidesteps the issue.

dbuenzli:
Here are two example of packages that provides libraries that use C a simple one:

https://github.com/dbuenzli/mtime/blob/master/myocamlbuild.ml
https://github.com/dbuenzli/mtime/blob/master/_tags

and tsdl whose pkg-config support has been at least c&p once by @Armael: 

https://github.com/dbuenzli/tsdl/blob/master/_tags
https://github.com/dbuenzli/tsdl/blob/master/myocamlbuild.ml

gasche:
Thanks! Those examples are very useful.

Through personal experimentations with `link_c_lib` implementation attempts, observation of C+OCaml libraries in the wild, plus looking at @dbuenzli's examples, I have determined that I unqualified to write solid rules for C code compilation, and that I should leave this to someone that knows better. I'm glad to help where I can on the ocamlbuild front, but I don't have the time, right now, to learn about whether `-cclib libtest.a` or `-cclib -ltest` is best, or why @dbuenzli uses `-dllib` instead of `-cclib` for (custom)) bytecode compilation.

I think this might be an opportunity for an external plugin to shine (in particular, such a plugin would be accessible to users of previous OCamlbuild versions), but I would be equally happy merging suggestions into the OCamlbuild codebase proper. What matters is the goal: I would like ocamlbuild to make it as easy, for beginners, to build mixed C/OCaml programs, than it currently makes building pure-OCaml projects.

@Armael , would you maybe be interested in pluginfying @dbuenzli's pkg-config code, for a start?

whitequark:
To answer these particular questions for anyone else: `-cclib -ltest` is proper, since that allows builds on systems that don't use the `.a` extension, i.e. Windows (in case MSVC is used, it would be the job of ocamlc to translate `-lfoo` to `libfoo.lib`, which you can't robustly do with a full filename); and `-dllib` records a `.so` to be loaded by the bytecode toplevel, whereas `-cclib` records a flag to be passed to the system linker by the bytecode compiler.

gasche:
Another question I had: is `lib` suffix a convention for C libraries? When I looked at Zarith for example, it embeds `-lzarith` in its `.cma` file but distributes `zarith.a` and not `libzarith.a`.

whitequark:
I'm surprised that works (maybe GNU linker is permissive here?). Yeah, the `lib` prefix is the convention.

dbuenzli:
Also note that the section [§19.1](http://caml.inria.fr/pub/docs/manual-ocaml/intfc.html#sec414) of the OCaml manual has most of the information one needs to know.

Issue #32, opened 2015-12-26
We need a release in opam


gasche:
Given that ocamlbuild development started here at an enthusiastic pace, I'm about to send a pull request to ocaml/ocaml that removes ocamlbuild. However, doing this right now would break build for people using 4.03+dev, because the only ocamlbuild package available on opam is the phony one relying on the compiler distribution.

Of course ocamlbuild developers can easily pin an opam file (well as soon as we have one, confer #29 ), but 4.03+dev testers should not need to pin ocamlbuild to preserve their current and future release-testing workflow.

If I understand correctly, this means that we should try to have some release on OPAM, packaged as working on 4.03+dev but not on previous versions. This would not need to be a production-quality release: it's for 4.03+dev testers so it can be bleeding edge.

This raises two questions:
1. what should we release?
2. will there be a transition period for people using the current ocaml trunk (with ocamlbuild included) and installing our new release before updating their compiler trunk version?

Point 2. might be an invitation for the interested contributor to experiment with what happens when someone installs a non-phony ocamlbuild version (such as our coming release) on top of an ocaml switch that does distribute ocamlbuild by default. Maybe we can make it work (which would be nice to let 4.02.3 users use our shiny future ocamlbuild releases), or maybe we need to detect this case and defensively refuse to build, etc.

On point (1.): we could release ocamlbuild:
- as at was at the 4.02.3 release
- as it is about now
- as it is after #5 is merged
- after all ['split-project-start' issues](https://github.com/ocaml/ocamlbuild/labels/split-project-start) have been resolved

I would personally aim for "as soon as possible to get the transition done", but I'm open to all opinions on this.

gasche:
For the record, my ocamlbuild-less compiler branch is [gasche/ocaml#remove-ocamlbuild](https://github.com/gasche/ocaml/tree/remove-ocamlbuild), which can be used as a starting point to experiment with the post-removal transition.

gasche:
I submitted the 0.9.0 release in OPAM:

  https://github.com/ocaml/opam-repository/pull/5483

gasche:
0.9.0 was merged in OPAM, so this can be closed.

PR #38, opened 2016-01-01
An opam-ready build system?

gasche:
This PR builds upon @rgrinberg's #29 to provide a hopefully-satisfying installation procedure for ocamlbuild. In the light of #32, I think it is important to get something "releasable in OPAM" as soon as possible, and my hope is that this PR (with modification after review etc.) gives us a "releasable" `Makefile + opam` combination.

The design constraints imposed on the proposed Makefile are:
- it can be used without opam: `make LIBDIR=$(ocamlc -where)/.. BINDIR=/usr/local/bin install uninstall` should work
- it does not require a hard-dependency on ocamlfind (cc @avsm), as it installs the META file itself
- it works in environments that do not have a native compiler available (`make NATIVE=false`)

I incrementally designed three different kinds of install/uninstall targets:
- `install uninstall`, which does (un)installation manually, no findlib dependency (a potential choice for distribution packagers)
- `findlib-install finlib-uninstall`, which uses `ocamlfind {install,remove}` (except for the binaries)
- `opam-install` which does not actually install anything but merely generates an `ocamlbuild.install` file (the choice of scripting the `.install` file generation comes from @dbuenzli's advice in #29 and ocaml/opam#2392)

I tested that the `install uninstall` roundtrip works with and without a native compiler available.

This packaging is not designed to be used with an ocaml distribution that ships ocamlbuild already. The current behavior is that the install targets will override the installed ocamlbuild (binaries and libraries), but that the uninstall targets do not restore the dummy ocamlfind package (then you need to `opam reinstall findlib` to un-break the environment). I don't know for sure whether it is possible and reasonable to try to support this use-case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@
 *.out
 *.out.dSYM
 
+ocamlbuild.install
+
 /src/ocamlbuild_config.ml
 /src/lexers.ml
 /src/glob_lexer.ml
--- a/META
+++ b/META
@@ -1,0 +1,7 @@
+# Specification for the "ocamlbuild" library
+requires = "unix"
+version = "[distributed with Ocaml]"
+description = "ocamlbuild support library"
+directory= "^"
+archive(byte) = "ocamlbuildlib.cma"
+archive(native) = "ocamlbuildlib.cmxa"
--- a/META
+++ b/META
@@ -2,6 +2,5 @@
 requires = "unix"
 version = "0.9"
 description = "ocamlbuild support library"
-directory= "^"
 archive(byte) = "ocamlbuildlib.cma"
 archive(native) = "ocamlbuildlib.cmxa"
--- a/META
+++ b/META
@@ -1,6 +1,6 @@
 # Specification for the "ocamlbuild" library
 requires = "unix"
-version = "[distributed with Ocaml]"
+version = "0.9"
 description = "ocamlbuild support library"
 directory= "^"
 archive(byte) = "ocamlbuildlib.cma"
--- a/Makefile
+++ b/Makefile
@@ -11,16 +11,28 @@
 #                                                                       #
 #########################################################################
 
-OCAMLC    = ocamlc
-OCAMLOPT  = ocamlopt
-OCAMLDEP  = ocamldep
-OCAMLLEX  = ocamllex
-CP        = cp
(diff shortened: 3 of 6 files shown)


gasche:
I just added a check that aborts the installation if the OCaml compiler already has a preinstalled ocamlbuild version (unless `CHECK_NOT_PREINSTALLED=false` is passed in the environment). This should make it much harder for an opam release of this version to make tricky-to-recover-from changes to user's environments.

gasche:
I experimented some more and I now suspect that a proper `./configure` step would help simplify the build, and make it easier to make changes in the future. In particular, I dislike having to include `$(ocamlc -where)/Makefile.config` from the main Makefile, as it may grow in the future and shadow other build variables.

I'll try to update this PR to move to a separate `./configure` step as soon as possible.

gasche:
@bobot excellent question, I also wondered about that. In #41 I chose GNU make. Feedback is warmly welcome.

(I did not force-push the present PR because I'm not sure yet that #41 is clearly the better choice.)

Hopefully a separate configure step is also a good start to solve the show-stopper bug #40.

adrien-n:
Do you want this to build with a POSIX make or are GNU-isms OK?

gasche:
I *think* that the `ifeq (...)` that I used in #41are GNU-make-only. I am fine with using GNU make if there is a substantial gain in simplicity (BSD people repeatedly claimed that this was not a problem for them in practice; I don't know but I actually suspect that opam's `make` resolves to `gmake` by default, at least grepping for `gmake` in opam-repository does not find anything). If if it is easy to make the Makefile portable at a low readability cost, I of course welcome patch proposal to do so. My main worry is actually to also have it work on Windows (I trust Make to work well, but whenever I write shell scripts I worry about Windows, and I don't have a machine to test).

adrien-n:
My question was instead aimed at adding more GNU-isms. :)

gasche:
No problem with this in principle. But note that I'm interested in moving relatively fast on this: as explained in #32, we need to be ready to make a release as soon as possible, and that means killing as many of the open [split-project-start](https://github.com/ocaml/ocamlbuild/labels/split-project-start) issues as possible. I will be happy to receive PRs to make the build system jaw-dropping, but a solid second opinion on "this one of the two proposals feels mergeable" in a shorter time frame is my primary goal.

gasche:
Superseded by the merge of #41.

Issue #13, opened 2015-12-22
main API should not export utility modules


agarwal:
The main API defines modules like `List` and `String`. This is not a high priority issue, but I feel this is inappropriate and should be removed. The particular (easy to workaround) problem I had is:

```ocaml
module List = struct
  include List
  include ListLabels
end
open Ocamlbuild_plugin
(* Oops, my List module above is shadowed. *)
```

gasche:
This is a fair point. We should move these modules to an `Library` submodule of the `PLUGIN` interface, so that people previously using these functions can still access them easily.

Another nice thing to do would be to expose modules with only the new functions added by ocamlbuild, in for example `Library.ListExtension` (so `Library.List` would just be `struct include List include ListExtension end`), so that people can open or include `ListExtension` to add to their own modules if they find its features useful.

agarwal:
My solution to this is:

- All modules are namespace'd by using a prefix, so here module names would be OCamlbuild_tags, OCamlbuild_list, etc.
- Define a `Std` module (or name it whatever you want). This is the main API entry point. Within this module, you export the modules that are the *real* API.

Now, all modules are accessible if somebody really wants them, but it's clear that some aren't really what the project provides.

rgrinberg:
+1 for @agarwal suggestion

PR #44, opened 2016-01-14
Better cycle refinement computation

bobot:
Currently when circular dependencies are detected the following is printed:
```
Circular build detected
  (src/types.cmx already seen in [ src/types.cmx; src/explanation.cmx;
                                   src/conflict.cmx; src/cmd/popop.cmx;
                                   src/cmd/popop.native ])
```

In this case `Types` was wrongly used in `types.ml`.  It is more informative to have just the cycle:

```
Circular build detected (src/types.cmx already seen in [ src/types.cmx ])
```

This merge request refine the witness by keeping only the cycle. It also simplify a similar refinement done when there is a cycle during the computation of the ocaml files to link.

I just don't know where to put this `refine_cycle` function. At first I wanted to put it in `My_Std.List` but it seems that it should not (cf. #13). So, where should I put it?

### DIFF
--- a/src/ocaml_dependencies.ml
+++ b/src/ocaml_dependencies.ml
@@ -195,66 +195,37 @@ module Make (I : INPUT) = struct
     let dependencies_of x =
       try SMap.find x !*dependencies with Not_found -> Resources.empty in
 
-    let refine_cycle files starting_file =
-      (* We are looking for a cycle starting from [fn], included in
-         [files]; we'll simply use a DFS which builds a path until it
-         finds a circularity.
-
-         Note that if there is at least one cycle going through [fn],
-         calling [dfs path fn] will return it no matter what [path] is
-         (it may just not be the shortest possible cycle). This means
-         that if [dfs path fn] returns [None], [fn] is a dead-end that
-         should never be explored again.
-       *)
-      let dead_ends = ref Resources.empty in
-      let rec dfs path fn =
-        let through_dep f = function
-          | Some _ as cycle -> cycle
-          | None ->
-             if List.mem f path
-             then (* we have found a cycle *)
-               Some (List.rev path)
-             else if not (Resources.mem f files)
-             then
-               (* the neighbor is not in the set of paths known to have a cycle *)
-               None
-             else
-               (* look for cycles going through this neighbor *)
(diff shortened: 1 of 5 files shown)


gasche:
The patch seems correct and it's a good simplification, but I dislike the fact that the new `refine_cycle` is defined twice. Is it not possible to declare it only once?

gasche:
Or you could just implement

```ocaml
val index_of : 'a -> 'a list -> int option
val split_at : int -> 'a list -> 'a list * 'a list
```

in `My_std.List` and use those directly.

src/ocaml_dependencies.ml
```diff
@@ -195,66 +195,30 @@ module Make (I : INPUT) = struct
...
-        | None -> Resources.elements files
-        | Some cycle -> cycle
-    in
-
     let needed_in_order = ref [] in
     let needed = ref Resources.empty in
-    let seen = ref Resources.empty in
-    let rec aux fn =
+    let rec aux on_the_go fn =
       if sys_file_exists fn && not (Resources.mem fn !needed) then begin
```
gasche: I feel a bit lost re-reading this code after a while. Why is it testing `sys_file_exists fn` here, and only adding the files of `(dependencies_of ...)` if they do *not* exists below?

bobot: I don't understand what you are referring as _adding the files_.  Nothing is done with file that don't exists, the last `else` concern the _second_ `if` ,`Filename.check_suffix f ".cmi"`, not the first, `sys_file_exists f`.

gasche:
I would feel more confident if someone else could review this PR. @Armael , would you maybe feel motivated?

Armael:
Sure. I dropped a few comments; that's probably me just being stupid, i'm trying to figure out what's going on. The last commit also probably invalidates my 2nd comment, but with a commit message that seems incompatible with what i deduced from the second commit... I'm confused.

Armael:
Ok! Looks good to me then.

gasche:
I'll trust that opinion and merge the PR. @bobot , could you please add a mention of your work in the Changes file?

gasche:
Thanks!

PR #51, opened 2016-02-10
revert commit 1d8e590 from OCaml (PR#6475)

damiendoligez:
This reverts the part of ocaml/ocaml@1d8e590c54dcd888503c3ea2944533255ef87343 that was in ocamlbuild.

It's needed because the corresponding feature is (temporarily) removed from OCaml.

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -536,7 +536,9 @@ rule "ocaml C stubs: c -> o"
     let c = env "%.c" in
     let o = env x_o in
     let comp = if Tags.mem "native" (tags_of_pathname c) then !Options.ocamlopt else !Options.ocamlc in
-    Cmd(S[comp; T(tags_of_pathname c++"c"++"compile"); A"-c"; A"-o"; P o; Px c])
+    let cc = Cmd(S[comp; T(tags_of_pathname c++"c"++"compile"); A"-c"; Px c]) in
+    if Pathname.dirname o = Pathname.current_dir_name then cc
+    else Seq[cc; mv (Pathname.basename o) o]
   end;;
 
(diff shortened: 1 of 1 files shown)


damiendoligez:
See also ocaml/ocaml#464.

gasche:
@whitequark, you are in charge :-)

whitequark:
Sure.

gasche:
@bobot thinking about this more, I'm not sure why we need to do a release quickly, though. Can't you just pin ocamlbuild to its master branch while doing your testing?

The reason I ask instead of just releasing is that I expect this situation to happen again in the future: OCaml upstream *will* change its compilation behaviour in ways that require ocamlbuild changes. I'm not sure we should immediately do a release -- at least if there is an another way to use some "safe with trunk" branch without constantly making releases.

In this specific case, we know this behavior will be as currently in the 4.03 release, and we don't expect other ocaml-side changes to break ocamlbuild, so it makes sense to have a release between now and 4.03 release anyway. So I'm fine with having a release. But I would like you (or someone else) to work out a process to use pinning instead, because in the general case we shouldn't have to release to keep in synch.

gasche:
@whitequark , now that ocaml/ocaml#464 has been merged, do you have any advice on how to handle it on the ocamlbuild side? Given that people may use recent ocamlbuild version but keep using 4.03, we probably can't just revert the revert; I suspect that the easiest way is to leave the code as-is, maybe with a comment there that >=4.04 OCaml version would actually support the `-o` option, to revisit later when we don't need to support 4.03 anymore.

whitequark:
@gasche It is rather unfortunate, but I agree that there's probably no other way forward.

PR #53, opened 2016-02-19
a tentative Release howto guide -- a bit long

gasche:
@bobot this would be my reply (too long as usual) to [you wondering](https://github.com/ocaml/ocamlbuild/pull/51#issuecomment-186163361) about a release guide.

### DIFF
--- a/howto/release.adoc
+++ b/howto/release.adoc
@@ -1,0 +1,202 @@
+= How to release a new OCamlbuild version =
+
+The steps below describe what needs to be done to release a new
+OCamlbuild version. The process should be simple and fast, so that we
+can release often.
+
+== Preparing the release ==
+
+The steps can be done a few days in advance, and coordinated among
+several people. External contributors can do it, and you can use pull
+requests to get reviews and feedback.
+
+[[synch-opam-files]]
+=== Synchronizing the local and remote opam files ===
+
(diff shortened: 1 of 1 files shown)


gasche:
(Some part could be extracted to an independent CONTRIBUTING.md document, but let's focus on the need for a release first.)

Changes
```diff
@@ -23,6 +23,8 @@ NEXT_RELEASE:
 OCamlbuild 0.9.0 (19 Jan 2016):
 -------------------------------
 
+- #44: When ocamlbuild discover a cycle in the dependencies, the
```
hcarty: Should be "discovers"

gasche: Fixed, thanks. (Github seems confused about which patches are part of this PR, this one actually isn't, but that's fine)

gasche:
In 9ebb6146f38b455f5a8c1c7bddace49fad71d5b9 I added a release summary. @bobot , if you agree with the steps proposed in the release summary, you should be able to do the rest of the release work. I think it would be a good test.

gasche:
I realize that the howto does not mention that the version number is written in some places of the repo and needs to be updated. `git grep -F "0.9"` is a good way to find those places; currently `META` and `VERSION` need to be updated at each release. I should update the howto.

PR #56, opened 2016-02-24
Readme

bobot:
With information about opam pinning

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -4,14 +4,14 @@
 OCamlbuild is a generic build tool, that have built-in rules for
 building OCaml library and programs.
 
-Your should refer to the (OCambuild
-manual)[https://github.com/gasche/manual-ocamlbuild/blob/master/manual.md]
+Your should refer to the [OCambuild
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -0,0 +1,26 @@
...
+OCamlbuild is a generic build tool, that have built-in rules for
+building OCaml library and programs.
+
+Your should refer to the [OCambuild
+manual](https://github.com/gasche/manual-ocamlbuild/blob/master/manual.md)
+for more informations on the tools.
+
+## Automatic Installation ##
+
+OCamlbuild was directly present in OCaml release prior to 4.3. Now it
```
gasche: 4.03 seems to be the more common spelling of the name. Maybe you could point to [the caml-list announce](https://sympa.inria.fr/sympa/arc/caml-list/2015-12/msg00003.html) email that has more details on the split? We could have for example:

```
> Until OCaml version 4.02, OCamlbuild was distributed as part of the compiler distribution. [Since OCaml 4.03](https://sympa.inria.fr/sympa/arc/caml-list/2015-12/msg00003.html), it is now distributed separately.
```

PR #63, opened 2016-03-04
Added simple install instructions.

bschommer:
First try for simple compile instructions for ocamlbuild without
opam.

The step with the creation of the dummy depend file is only needed to the Makefile from bailing out. Another solution would be to use -include instead of include in the Makefile.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -26,30 +26,21 @@ opam pin add ocamlbuild --kind=git "https://github.com/ocaml/ocamlbuild.git#mast
 
 ## Compilation from source ##
 
-1. Configure the system.
-
+1. Configure.
 ```
-make -f configure.make Makefile.config OCAMLBUILD_PREFIX=$prefix 
-         OCAMLBUILD_BINDIR=$prefix/bin OCAMLBUILD_LIBDIR=$prefix/lib/ocaml 
-         OCAML_NATIVE=$native OCAML_NATIVE_TOOLS=$native
+make configure
 ```
 
-where ```$native``` is true if native tools are available and ```$prefix``` is
-the intended installation lcoation
+The installation location is determined by the installation location of the ocaml
+compiler. For installation in a different location just change the paths in the
+generated Makefile.config.
 
-2. Create the dependency file.
-
+2. Compile the sources.
 ```
-touch .depend
-make depend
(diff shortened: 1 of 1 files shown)


gasche:
I'm far from a `make` expert and didn't know about `-include`, thanks for the tip! However, I'm surprised that `.depend` would be a problem, as it is versioned with the repository. (I just checked that it is also present in the archive we distributed.) Aren't you rather thinking of the `Makefile.config` file?

On my machine, something rather magical happens: just typing `make` in *absence* of a Makefile.config file works perfectly. Looking at gmake's documentation, it seems that this is a feature: when an `include`d file is missing, make will [try to remake it](https://www.gnu.org/software/make/manual/html_node/Remaking-Makefiles.html#Remaking-Makefiles). So upon finding `include Makefile.config`, my make runs its `Makefile.config` rule that does the right thing. If I comment the section of the main Makefile named "the config file", then this does not work anymore.

This is of course a bit too clever for its own good. Questions:
- what's a good way to make the script less clever? One would be to require users to explicitly do the configure step first (`make -f configure.make ...`), as in the proposed PR. Another would be to use `-include` as suggested, but do `make configure` first (I intended this to be the user-visible entry point for configuration, but maybe that's not such a good idea).
- what about portability? In my mind, our makefile generally requires GNU make (in particular I wouldn't expect the conditionals or functions used to work on other implementations). @bschommer, does it fail on your machine because you use another make implementation? Shouldn't we explicitly ask people to use GNU make to avoid tedious portability concerns?

A partially-related remark on your instructions: there some initialization logic in `configure.make` to make sure that reasonable variables are chosen by default (it tries to use `opam` to get good defaults, then `ocamlfind`, then the ocaml compiler's variables). It *should* work out-of-the-box for most users. Is it the case for you, @bschommer? If it does work, maybe we could rather show the simple `make -f configure.make` or `make configure` call in the instructions, and as a second step list the `OCAMLBUILD_*` and `OCAML_*` as *optional* configuration knobs.

bschommer:
I'm also not a large ```make``` expert.

You are right just typing ```make`` already works perfectly fine for me. I suspect my problems came from the fact that I had several different ocaml version on my path.

PR #68, opened 2016-04-06
Add support for flambda's optimization options to OCamlbuild

martin-neuhaeusser:
Add support for the new optimization command line options brought by flambda.
The OCamlbuild flags are named according to the flambda options with two exceptions (to clarify the meaning of the flag):

* `-rounds <int>` can be specified with the `optimization_rounds(int)` flag
* `-O{2,3,classic}` can be specified with the `optimize({2,3,classic})` flag

### DIFF
--- a/Changes
+++ b/Changes
@@ -29,6 +29,8 @@ branch; this version that should be usable for testing packages using
   (François Bobot, review by Armaël Géneau)
 - #51: synchronize handling of -o flag for C files with upstream OCaml
   (whitequark)
+- #67: Add support for flambda's optimization parameters
+  (martin-neuhaeusser)
 
 OCamlbuild 0.9.0 (19 Jan 2016):
 -------------------------------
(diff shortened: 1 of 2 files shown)


gasche:
Does your choice of documentation parameter reflect the current default values? Will we want to keep them in sync' in the future?

gasche:
The patch is fine as far as I'm concerned, I'll merge it later if there is no other comments.

Just, could you cut the lines above 80 columns? (One certainly is, I can't browse it in the github diff, I suspect some of the `flags` call may be as well). We should check that it is documented, but it's better to keep respecting this old rule.

martin-neuhaeusser:
Yes, the default values correspond to those documented [here](https://ocaml.janestreet.com/ocaml-core/flambda_manual/#sec450).
I will adjust the overlong columns in a minute.

gasche:
While you're at it, you should also add a `Changes` entry. I just pushed a new section for the upcoming release, you'll see it if you rebase on your end.

martin-neuhaeusser:
Did the changes and rebased into a single commit.

gasche:
The Changes option should go in the next section. I'll cleanup myself (I've been nitpicky enough already), add an information on option naming in the Changes entry (it also serves as user documentation), and update the manual. Thanks!

PR #70, opened 2016-04-12
quick windows fixes for 4.03

fdopen:
three windows related changes - feel free to cherry-pick:

1) 
`My_unix.run_and_open` will either call the internal My_unix.run_and_open (https://github.com/ocaml/ocamlbuild/blob/982fdf1b9ecf1448f07aa74a43e07db8c410dbd6/src/my_unix.ml#L62) or Ocamlbuild_unix_plugin.run_and_open (https://github.com/ocaml/ocamlbuild/blob/982fdf1b9ecf1448f07aa74a43e07db8c410dbd6/src/ocamlbuild_unix_plugin.ml#L50).

The first function will use bash, the second plain cmd.exe. That's confusing. You can't always know, which function will be called, but you have to quote differently for both cases (`Filename.quote` vs `Shell.quote_filename_if_needed`). It can happen e.g. here: https://github.com/ocaml/ocamlbuild/blob/982fdf1b9ecf1448f07aa74a43e07db8c410dbd6/src/ocaml_specific.ml#L752 `2>/dev/null` is obviously wrong for cmd.exe, but bash.exe will accept it.

2)
Proper shell quoting. The code is copy&pasted from the ocaml source tree, because it is not exported. (See http://caml.inria.fr/mantis/view.php?id=6287 )

3)
Another side effect of allowing file names that starts with '_'. `Hygiene.check` now complains frequently about files under '_build/' (eg. myocamlbuild.cmi). The default build dir is: `let build_dir = ref (Filename.concat (Sys.getcwd ()) "_build")`. `Filename.concat` always use backslashes as path separator. `My_std.filename_concat` (`/`) however uses slashes, therefore the comparison here ( https://github.com/ocaml/ocamlbuild/blob/13feb3cafc1f4f664aca0a744f2f08e30bffd06a/src/main.ml#L162 ) won't return the desired result. (see https://github.com/ocaml/opam-repository/issues/6224 )

I've just changed `Pathname.normalize`, that previously didn't work reliable on windows at all, because only slashes were treated as path separators. It now always return a version of the file name with an uppercase drive letter and only forward slashes. (I've first tried to return a string with backslashes, but the solver didn't like it.)

### DIFF
--- a/src/ocamlbuild_unix_plugin.ml
+++ b/src/ocamlbuild_unix_plugin.ml
@@ -48,6 +48,14 @@ let at_exit_once callback =
   end
 
 let run_and_open s kont =
+  let s =
+    (* Be consistent! My_unix.run_and_open uses My_std.sys_command and
+       sys_command uses bash. *)
+    if Sys.win32 then
+      "bash --norc -c " ^ Filename.quote s
+    else
+      s
+  in
   let ic = Unix.open_process_in s in
   let close () =
     match Unix.close_process_in ic with
--- a/src/shell.ml
+++ b/src/shell.ml
@@ -24,12 +24,27 @@ let is_simple_filename s =
     | 'a'..'z' | 'A'..'Z' | '0'..'9' | '.' | '-' | '/' | '_' | ':' | '@' | '+' | ',' -> loop (pos + 1)
     | _ -> false in
   loop 0
+
+
+(* generic_quote and unix_quote are copy&pasted from
+   Ocaml source stdlib/filename.ml.
+   See also mantis issue #6287 *)
+let generic_quote quotequote s =
+  let l = String.length s in
+  let b = Buffer.create (l + 20) in
+  Buffer.add_char b '\'';
+  for i = 0 to l - 1 do
+    if s.[i] = '\''
+    then Buffer.add_string b quotequote
+    else Buffer.add_char b  s.[i]
+  done;
+  Buffer.add_char b '\'';
+  Buffer.contents b
+let unix_quote = generic_quote "'\\''"
+
 let quote_filename_if_needed s =
   if is_simple_filename s then s
-  (* We should probably be using [Filename.unix_quote] except that function
-   * isn't exported. Users on Windows will have to live with not being able to
-   * install OCaml into c:\o'caml. Too bad. *)
-  else if Sys.os_type = "Win32" then Printf.sprintf "'%s'" s
+  else if Sys.os_type = "Win32" then unix_quote s
   else Filename.quote s
 let chdir dir =
   reset_filesys_cache ();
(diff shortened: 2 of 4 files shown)


src/ocamlbuild_unix_plugin.ml
```diff
@@ -48,6 +48,14 @@ let at_exit_once callback =
...
 
 let run_and_open s kont =
+  let s =
+    (* Be consistent! My_unix.run_and_open uses My_std.sys_command and
+       sys_command uses bash. *)
+    if Sys.win32 then
+      "bash --norc -c " ^ Filename.quote s
+    else
+      s
+  in
```
gasche: Is there a reason for not calling `sys_command` directly?

src/shell.ml
```diff
@@ -24,12 +24,23 @@ let is_simple_filename s =
     | 'a'..'z' | 'A'..'Z' | '0'..'9' | '.' | '-' | '/' | '_' | ':' | '@' | '+' | ',' -> loop (pos + 1)
     | _ -> false in
   loop 0
+
+let generic_quote quotequote s =
```
gasche: could you include a comment tracing back the origin of the source code here?

gasche:
I don't know anything about windows programming so I won't be able to review this PR. @whitequark or @dra27, any suggestions?

src/ocamlbuild_unix_plugin.ml
fdopen: Why is ` implem.run_and_open` overwritten in the first place?
I've assumed to avoid a temporary file, if `Unix.open_process_in` is used instead of `Sys.command`.

gasche:
@fdopen, I would like to make a release for the new 4.03-specific flags, and I am not sure which portions of this PR to include in it. In the end, I would like to include everything, but I haven't had any feedback from other people using ocamlbuild on Windows, so I won't include the more invasive changes right now.

I wrote a smaller patch (pushed in master) trying to fix the detection of the build directory in the hygiene check, b29023c86813a13dded7cece90d829bb9c8f5372 . Could you maybe give it a try and say if it suffices to solve the problem (3) on your machine?

msprotz:
1) and 2) are good but I would like to see 2) fixed properly, a.k.a. export unix_quote in the standard library; just hide it after (** / **) so that it doesn't appear in the doc if it's controversial

Please be aware that there's a MASSIVE caveat with running `bash -c` (see http://stackoverflow.com/questions/9946586/cygwin-bash-c-yields-different-argc-depending-on-whether-the-command-start-with) and that

hhugo:
(1) was included in #329
(2) is part of #333 
(3) probably no longer needed now that we have b29023c86813a13dded7cece90d829bb9c8f5372

PR #74, opened 2016-04-21
ocamlfind ocamldep supports -predicates since version 1.5

diml:


gasche:
How should we publish our dependency on specific `ocamlfind` versions? Would adding `ocamlfind < "1.5"` as a `conflicts` clause work? If so, could you add this to the pull request?

diml:
The `conflits` clause should do, I added it

gasche:
Debian stable only has ocamlfind 1.4.1 packaged. Is that an issue? I assume that separate-ocamlbuild would not packaged before the next stable anyway, and testing already uses 1.5.2. So I guess it's fine.

PR #75, opened 2016-04-22
Multi parametrized tags

diml:
This patch adds support for parametrized tags where all the parameters for a given target are processed at once instead of one by one.

## Motivation

The use case I have for this feature is the following: in ppx\_driver I have an [ocamlbuild plugin](https://github.com/janestreet/ppx_driver/blob/master/ocamlbuild_plugin/ppx_driver_ocamlbuild.ml) to make it easier to build custom drivers for a given set of rewriters. Users just have to write this in their `_tags` file:

```ocaml
<src/*.ml>: ppx-driver(ppx_sexp_conv+ppx_inline_test+ppx_expect_test)
```

Then the plugin builds a custom driver for this set of rewriters and pass it as one single `-ppx` (in fact it's currently passed as `-pp` but it's not very important).

The only annoying thing with this is that this list is often duplicated in `_oasis` or somewhere else, as these rewriters must be specificed as runtime dependencies anyway (for the various supporting runtime libraries). What I'd  like instead is to have the same experience as with regular use of ppx where findlib implicitly adds many `-ppx` arguments.

With this feature I can reach this goal: the user just has to tag everything with a special tag (for instance `auto_ppx_driver`) and the ppx\_driver plugin will take care of the rest by using `pflag_multi`:

```
pflag_multi ["ocaml"; "compile"] "package" (fun packages -> ...)
```

I also wanted this several time in the past for other use cases.

## Implementation

Multi parametrized tag actions are kept in a separate table in `Param_tags`. `Param_tags.init` still calls the associated actions for every singleton, to make sure the check for used tags still works properly.

Then `Command.deps_of_tags` calls `Param_tags.handle_multi_param_tags` which makes sure that the actions have been called for this given set of parameters. I assume it's enough as `deps_of_tags` is always called first when building a target.

I added tests in `testsuite/internal.ml`


gasche:
Thanks! I haven't looked at the code yet, but if I understand correctly you want tags with a collecting semantics (we see all arguments that are passed over several separate use on the same target). To me the name "multi parametrized" rather suggests a tag with multiple parameter, so we should probably change it.

If today you try to use usual tags in this fashion, you get strange behaviours depending on how the handler is implemented. (You could in particular try to approximate the collecting semantics for a specific tag by keeping global state around, but that would be very messy: first it's unclear how to have robust per-target state, and second each successive application would have to undo the global effect of the previous one to re-do the effects of the new collection.)

If we re-did the system from scratch, it would probably be rather natural to give to *all* parametrized tags this collecting semantics, wouldn't it? The reason why you cannot do this now is that the interface of parametrized tag handlers has to change (take a list of parameters instead of one parameter), which would break existing code. But long-term does that mean that we should deprecate the non-collecting version, and encourage users to move to the collecting versions? (That could factor in the naming choice: if it's the right default it should be the most convenient to use.)

Is there a more flexible interface that could be used to have a better extensibility story in the future? For example I could see an argument being either `Single_use of (string -> unit) | Collecting of (string list -> unit)`, or something like that, which would allow to add richer interfaces in the future. That sounds overkill.

When is the decision made that all choices have been collected? I suppose this corresponds to when the tags are "acknowledged"; but doesn't this mean that we have to be a bit more careful? (Previously we could call the acknolwedgment several times if we had forgotten some parametrized tags, now that option is not available anymore: we need a well-defined stable point where no more tags will be added.)

diml:
Yh, the naming is not very good indeed. What about `pflag_collect`? The collecting semantic seems indeed like the right default. I'm just not sure about performances, with the current implementation collecting tags might be slightly slower than non-collecting ones.

> When is the decision made that all choices have been collected? I suppose this corresponds to when the tags are "acknowledged"; but doesn't this mean that we have to be a bit more careful? (Previously we could call the acknolwedgment several times if we had forgotten some parametrized tags, now that option is not available anymore: we need a well-defined stable point where no more tags will be added.)

Basically there is no point where we know that all possible sets of paremeters for a given tag have been collected. For every target when executing the rules, we call the actions with any new set of parameters discovered.

It's hard to know all the possible sets before executing the rules; you'd have to generate all possible combinations, which might be a lot. You could reduce the set by partitioning the globs, but that's a lot more work...

I suppose the current approach is fine as long as the actions don't call `tag*`, right? It's not clear what that mean anyway. We can probably have a global flag to prevent new tags to be added.

diml:
I thought a bit more about this. With the current PR it's almost possible to implement a clever findlib based plugin that would compute the right compiler flags once and for all instead of calling ocamlfind for every command. It's not yet possible with the proposed API as you need to collect 2 tags: `package` and `predicate`.

What would be nice would be an API similar to `Core.Param` or `cmdliner` where you could do:

```
flag (required "ocaml"
      +> required "compile"
      +> optional "auto_ppx_driver"
      +> many_params "package"
      +> many_params "predicate"
      +> nil)
  (fun (auto_ppx_driver : bool) (packages : StringSet.t) (predicates : StringSet.t) -> ...)
```

I'm proposing this but I don't think I'll have time to work on this and sort out all the subtleties...

BTW, currently you can see a tag `A(B)` as either a whole tag or a parametric tag. I think it makes thing a bit more complicated than they should. We could instead have only parametric tags, and a warning about unused tag arguments

gasche:
I think that a generic solution to the problem may be to let users
intercept the whole set of tags passed to a target, instead of just
a statically selected set of flags (with at most one
parametrized). This should be relatively easy to do, by adding a list
of user-specified functions of type `Tags.t -> Command.spec` to the
`Flags.of_tag` function.

We could then add a minimal API to let users say something like "I
want all the following tags to be present, with some of them maybe
parametrized, return the set of instances present", and then "give me
the parameter of the instance of this parametrized tag in the given
set". Your initial
proposal is the specific instance where you only require all instances
of a single parametrized flag -- and users can implement their own
refinement themselves (eg. may-be-present flags as you mention),
from a (hopefully) "maximally expressive" capability provided by
the core.

What do you think?

PR #81, opened 2016-05-04
Fix link to CONTRIBUTING.adoc from manual.

lehy:


gasche:
Thanks again; it's silly that I forgot to check for links when moving those files at merge time.

PR #82, opened 2016-05-06
Added flambda-related tags

nilsbecker:
in a little section; defaults for the values were copied from https://github.com/ocaml/ocamlbuild/pull/68 .


gasche:
Excellent, thanks.

PR #93, opened 2016-06-16
Add tags for the compiler flags -noassert and -unsafe.

fpottier:


gasche:
Thanks! After clicking the `merge` button I noticed that I should have asked for a Changes notice. I'll add it myself.

PR #94, opened 2016-06-18
Add cc parameterized flag

rgrinberg:
Corresponds to the -cc flag ocamlc/ocamlopt take

### DIFF
--- a/Changes
+++ b/Changes
@@ -27,6 +27,10 @@ NEXT_RELEASE:
 - #93: added "noassert" and "unsafe" flag for corresponding ocaml{c,opt} options
   (François Pottier)
 
+- #94: added "cc", "cclib", and "ccopt" flags which correspond to the their
+  respective ocaml{c,opt} options
+  (Rudi Grinberg)
+
 0.9.2 (1 May 2016):
 -------------------
 
--- a/manual/manual.adoc
+++ b/manual/manual.adoc
@@ -692,6 +692,9 @@ Feel free to look at the implementation and link:CONTRIBUTING.adoc[send a patch]
 * `dtypes`
 * `for-pack(PackModule)`
 * `inline(5)`
+* `cc(ccomp)`
+* `cclib(-llibname)`
+* `ccopt(option)`
 * `keep_locs`
 * `linkall`
 * `no_alias_deps`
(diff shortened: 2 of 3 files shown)


gasche:
Could you please add a mention in the manual (`manual/manual.adoc`, look for e.g. `inline` to get the compiler-options tags listing), and in Changes? I'm thinking of also adding `cclib` and `ccopt` while we're at it.

Changes
```diff
@@ -27,6 +27,9 @@ NEXT_RELEASE:
 - #93: added "noassert" and "unsafe" flag for corresponding ocaml{c,opt} options
   (François Pottier)
 
+- #94: added "cc", "cclib", and "ccopt" flags which correspond to the their
+  respective ocaml{c,opt} options
+
```
gasche: > (Rudi Grinberg)

?

rgrinberg: Done

gasche:
Thanks!

dbuenzli:
So I've been trying to use that without success because it's part of the manual. It would really be better to publish a manual that corresponds to the currently released ocamlbuild.

rgrinberg:
Would also be nice to release a version of ocamlbuild with these changes (among many others).

@dbuenzli , but since I remember that you maintain compatibility with your code on older OCaml as well, perhaps just having a manual of the latest release isn't enough. It would be good to annotate which tags/flags are only available in recent versions of ocamlbuild.

dbuenzli:
> with your code on older OCaml as well, perhaps just having a manual of the latest release isn't enough.

Let's not complicate this too much this would already be a good improvement given our limited ressources. My first reaction when I saw that it didn't work was to use a distribution-splitted `ocamlbuild` just to find that it didn't work there either.

> It would be good to annotate which tags/flags are only available in recent versions of ocamlbuild.

This seems more reasonable.

dbuenzli:
@gasche could we get a release with this, it will significantly improve the life of people dealing with c stubs /cc @pqwy @avsm

gasche:
Thanks for the request, @dbuenzli, it's helpful information. I put this in my short-term TODO list when you wrote the message but haven't had the occasion to pop it yet.

PR #95, opened 2016-06-18
add .merlin

rgrinberg:


gasche:
Thanks! (I'm not sure why `bootstrap` is in there but I'll move this directory soon anyway and may drop the line on this occasion.)

PR #99, opened 2016-07-06
Add bin-annot flag

rgrinberg:
Create .cmt & .cmti files when compiling ocamlbuild. Very helpful when using
with merlin (or ocp-index).

I assume there's no need to check for older functions for OCaml as those would
have a builtin version of ocamlbuild anyway.


gasche:
`-bin-annot` was added in 4.00.0. It feels like ages! Thanks.

PR #111, opened 2016-10-13
Add nostdlib tag/flag for ocamlc

edgemaster:


gasche:
Very good, thanks! Could you make the following minor changes?

- the commit message should say "for the compilers" (ocamlc suggests bytecode-only)
- I would like to have a Changes entry (see the `Changes` file)
- in the manual, you should add `(since NEXT_RELEASE)` (I will replace it with the version number at release time) so that users know what relies on recent ocamlbuild versions -- we haven't done that consistently in the past but it's a good habit to start.

edgemaster:
Changes completed as requested.

gasche:
Merged, thanks.

PR #115, opened 2016-10-24
Ignore node_modules

yunxing:
node_modules is a directory used in yarn (https://yarnpkg.com/) that contains
all the build artifacts of a package's dependencies (similar to _build). To make some ocaml packages installable by node_modules, we should just ignore everything in this directory, like what we did for .git and .svn.


jordwalke:
This will likely benefit people who use BuckleScript and js_of_ocaml. Thanks @yunxing!

gasche:
Thanks! I will try to harmonize the two places which seem to list different directory names to ignore.

PR #125, opened 2016-11-24
Add -toolchain option, forwarded to findlib

whitequark:
I need this for implementing https://github.com/dbuenzli/topkg/issues/86 robustly, as the environment variable manipulation requires ExtUnix to cover every case (since `OCAMLFIND_TOOLCHAIN=""` is not the same as `unset OCAMLFIND_TOOLCHAIN`, but ocamlbuild can simply not set it).

### DIFF
--- a/Changes
+++ b/Changes
@@ -20,6 +20,9 @@ NEXT_RELEASE:
 - #115: add `node_modules` to the list of directories ignored by default
   (Yunxing Dai)
 
+- added "-toolchain" option for corresponding ocamlfind option.
+  (whitequark)
+
 0.9.3 (6 Oct 2016):
 -------------------
 
--- a/src/options.ml
+++ b/src/options.ml
@@ -39,6 +39,7 @@ let nostdlib = ref false
 let use_menhir = ref false
 let catch_errors = ref true
 let use_ocamlfind = ref false
+let toolchain = ref ""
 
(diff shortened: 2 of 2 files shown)


whitequark:
@gasche can you please review this?

gasche:
My first reaction was that having a command-line option was weird, as other compiler-related options are general passed as flags. On the other hand, it seems that this option has a specific placement (if must be right after the `ocamlfind` command) which makes the usual flag-handling inappropriate, so it would have to be a hardcoded flag which is not much better. Finally, another idea would be to have a command-changing option `-ocamlfind "ocamlfind -toolchain ..."`, more general, but this wouldn't necessarily mesh well with the name-resolution logic (to be etried), so I think the interface you propose is a reasonable choice.

Could you add an entry to `Changes`?

whitequark:
> Could you add an entry to Changes?

I did...

gasche:
(I remember now that we should also add it to the manual. I'll probably do that after merging.)

whitequark:
> I remember now that we should also add it to the manual.

Yes. I thought about this but wasn't sure where to explain it.

gasche:
Merged. I documented it in 21826e070795bfd4db558c523904eb33f51e8e4e.

whitequark:
Thanks!

PR #126, opened 2016-11-25
Put spaces around `^` for compatibility with macros

OlivierNicole:
I'm working on a modified version of the compiler with macros, and I would like to install ocamlbuild and various other packages. However, macros introduce a small backward incompatibility. In the following expression:
```ocaml
"some string"^M.some_other_string
```
`^` is no longer parsed as an operator but as part the lifted path `^M.some_other_string`, causing an error.

Even though macros are only a prototype for now, merging this pull request would allow people to install `ocamlbuild` with my modified compiler without having to pin and modify the source.

### DIFF
--- a/src/command.ml
+++ b/src/command.ml
@@ -369,7 +369,7 @@ let digest =
   fun x ->
     match cmd x [] with
     | [x] -> x
-    | xs  -> Digest.string ("["^String.concat ";" xs^"]")
+    | xs  -> Digest.string ("[" ^ String.concat ";" xs ^ "]")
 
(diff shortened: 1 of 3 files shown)


gasche:
I just merged the patch, but I'm a bit skeptical about the choice of a syntax that breaks existing code.

OlivierNicole:
Yeah it is not ideal. We're certainly open to suggestions. Maybe `.^Module` would be better, but it is quite heavyweight and would look very bad in combination with record labels: `record..^Module.field`

PR #127, opened 2016-11-26
install the ocamlbuild.1 man page

gasche:


gasche:
I tested the (new) install scripts and they seem correct, but I would like to update the manpage before a merge.

gasche:
I updated the PR with code to auto-generate the list of options in the manpage from the more authoritative source that is `src/options.ml`.

gasche:
Oh, sorry for that and thanks for the fix.

Issue #129, opened 2016-11-29
-no-plugin semantics


samoht:
I am curious about the intended semantics of `-no-plugin`. Currently `-no-plugin` seems to imply "discard any pre-compiled `_build/myocamlbuild` binary". Is that what it was intended?

In some cases it is useful to separate the build of `_build/myocamlbuild` and the actual run of that command (for instance building + generating the doc without having to remember the full list of build plugins, see dbuenzli/topkg#80). How are we supposed to do that with `-no-plugin` and such?

Thanks!

gasche:
The specification of `-no-plugin` is that it will run plain ocamlbuild without being driven by a `myocamlbuild.ml` plugin -- it will apply the default rules and nothing else. I'm not sure whether you found a difference with respect to this specification.

If you want to build the plugin without building any targets, you can use the `-just-plugin` option. Finally, if you want to pass arguments to the moment where targets are built and not to the plugin build, you can use the (rather awkwardly named) `-plugin-option` (it's an option passed to the plugin, not to build the plugin like `-plugin-tag`).

gasche:
Note: the semantics of `-no-plugin`, `-just-plugin`, `-plugin-option` and `-plugin-tags` are explained [in the manual, section: How `myocamlbuild.ml` is used](https://github.com/ocaml/ocamlbuild/blob/21826e0/manual/manual.adoc#31-how-myocamlbuildml-is-used). @samoht, @whitequark, do you see any clarification that I should make to this documentation?

whitequark:
@gasche Yes. This passage:

> The -plugin-option FOO option will pass the command-line option FOO to the myocamlbuild invocation — and ignore it during plugin compilation.

makes it not clear what happens to FOO if there's no myocamlbuild.ml.

gasche:
What is the behavior that you would expect? Currently there is a warning if `-plugin-tags` is used but there is no plugin (because it's a common pitfall for people to assume that `-plugin-tag cppo.ocamlbuild` doesn't require any myocamlbuild.ml on their side), we could do the same for `-plugin-option`.

whitequark:
@gasche I would like the argument to -plugin-option to be used as if it was passed directly to ocamlbuild. Anything else would significantly complicate topkg.

gasche:
[...]

PR #130, opened 2016-11-30
make sure that -just-plugin always stops after the plugin-build phase

gasche:
Previously `-just-plugin` would only be checked at the end of plugin
build, and plugin build would only run if there is a plugin. This
means that `-just-plugin` would be ignored if there was no
`myocamlbuild.ml`: the command `ocamlbuild -just-plugin foo.byte`
 would build the target or not depending on whether a plugin source
was present.

This issue was reported by @whitequark in dbuenzli/topkg#87.

### DIFF
--- a/Changes
+++ b/Changes
@@ -26,6 +26,9 @@ NEXT_RELEASE:
 - #127: install ocamlbuild's man pages, missing since 4.02
   (Gabriel Scherer)
 
+- #130: make sure that -just-plugin always stops after the plugin-build phase
+  (Gabriel Scherer, report by whitequark)
+
 - #132: add a rule producing .cmxs from .mllib files.
   Previously, .mllib files would only produce .cm{x,}a files, and .cmxs were
   only produced by a separate (and in most cases redundant) .mldylib file.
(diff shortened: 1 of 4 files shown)


gasche:
See also #129 for a discussion of the `-no-plugin` and `-plugin-option` semantics.

whitequark:
@gasche ping?

gasche:
Well... I guess I can merge -- after fixing the conflicts.

whitequark:
Thanks a lot! When do you think I can expect a release of ocamlbuild?

gasche:
Well, whenever. Maybe in the next few days. Thanks for asking.

gasche:
(I'm preparing the release right now.)

gasche:
ocaml/opam-repository#8155

whitequark:
Many thanks!

whitequark:
@gasche I don't know how but this seems broken in master...

whitequark:
@gasche Oh, I see now. We misunderstood each other. In essence what I want is for ocamlbuild to not pass -toolchain when building a plugin, ever. If this is really desirable (which it never will be), -plugin-option is there.

whitequark:
So, https://github.com/ocaml/ocamlbuild/pull/160.

Issue #131, opened 2016-12-01
Add a .mllib -> .cmxs rule


dbuenzli:
This rule should be put after the `.mldylib` -> `.cmxs` rule so that it doesn't override a possible (but unlikely) different explicit way of building the shared library for a library.

See also discussion there https://github.com/dbuenzli/uuseg/pull/4

gasche:
We should have a testcase to make sure that the ordering is as we expect. Armael, this should go in `testsuite/internal.ml`, to be run from `testsuite/` as follows: `OCAMLBUILD=$(pwd)/../ocamlbuild.byte ocaml internal.ml`.

P.S.: Thanks a lot!

gasche:
Closed by merging #132. Thanks @dbuenzli and @Armael

gasche:
I'm suddenly confused: in 0.9.3 there was already a rule defined as follows

```ocaml
rule "ocaml: cmxa & a -> cmxs & so"
  ~prods:["%.cmxs"; x_dll]
  ~deps:["%.cmxa"; x_a]
  ~doc:"This rule allows to build a .cmxs from a .cmxa, to avoid having \
        to duplicate a .mllib file into a .mldylib."
  (Ocaml_compiler.native_shared_library_link ~tags:["linkall"] "%.cmxa" "%.cmxs");;
```

which is documented in the manual.

Why was a new rule necessary if this rule was present? Is there something that actually prevents it from working? Is there a conflict between this old rule and the new rule?

I am in the work of preparing a new release, and found out about this when giving finishing touches to the manual. I am considering reverting #132 from the release in order to give ourselves (that is, @Armael :-^ ) more time to think about this. Or it could still go in the release but be amended/removed if found redundant with the existing approach -- once this existing approach is fixed.

Armael:
I don't have any particular opinion about whether the rule added in #132 is actually needed or not. I also discovered the existence of the `cmxa -> cmxs` rule while implementing it, but thought for some reason a `mllib -> cmxs` rule was also needed, because of this issue.

gasche:
The following test case seems to work under 0.9.3:

```
let () = test "CmxsFromMllib"
  ~description:"Check that a .cmxs file can be built from a .mllib file"
  ~options:[`no_ocamlfind; `no_plugin]
  ~tree:[
    T.f "a.ml" ~content:"let a = 1\n";
    T.f "b.ml" ~content:"let b = true\n";
[...]

PR #132, opened 2016-12-01
Add a .mllib -> .cmxs rule

Armael:
Attempt at fixing #131 . Note that I'm a complete stranger to ocamlbuild rules — does this look like what one would expect?

### DIFF
--- a/Changes
+++ b/Changes
@@ -26,6 +26,13 @@ NEXT_RELEASE:
 - #127: install ocamlbuild's man pages, missing since 4.02
   (Gabriel Scherer)
 
+- #132: add a rule producing .cmxs from .mllib files.
+  Previously, .mllib files would only produce .cm{x,}a files, and .cmxs were
+  only produced by a separate (and in most cases redundant) .mldylib file.
+  This rule allows to produce .cmxs from .mllib files; .mldylib files
+  still take precedence when they exist.
+  (Armaël Guéneau, requested by Daniel Bünzli)
+
 0.9.3 (6 Oct 2016):
 -------------------
 
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -334,7 +334,7 @@ rule "ocaml: mllib & cmx* & o* -> cmxa & a"
         as the .cma rule above. Note that whereas bytecode .cma can \
         be used both for static and dynamic linking, .cmxa only support \
         static linking. For an archive usable with Dynlink, \
-        see the rule producing a .cmxs from a .mldylib."
+        see the rules producing a .cmxs from a .mllib or a .mldylib."
   (Ocaml_compiler.native_library_link_mllib "%.mllib" "%.cmxa");;
 
 rule "ocaml: p.cmx & p.o -> p.cmxa & p.a"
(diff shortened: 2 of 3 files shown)


Armael:
I added a test checking that the rule doesn't trigger if a .mldylib is present (and a Changes entry).

gasche:
(Did you check that the test does fail if you reverse the rule declaration order?)

Armael:
Yes.

More precisely, it fails on the following diff, or if I remove the .mldylib from the tree in the test case.
```
diff --git a/src/ocaml_specific.ml b/src/ocaml_specific.ml
index 627ba6d..bfa1b76 100644
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -351,18 +351,6 @@ rule "ocaml: cmx & o -> cmxa & a"
         transitivitely."
   (Ocaml_compiler.native_library_link "%.cmx" "%.cmxa");;

-rule "ocaml: mldylib & p.cmx* & p.o* -> p.cmxs & p.so"
-  ~prods:["%.p.cmxs"; x_p_dll]
-  ~dep:"%.mldylib"
-  (Ocaml_compiler.native_profile_shared_library_link_mldylib "%.mldylib" "%.p.cmxs");;
-
-rule "ocaml: mldylib & cmx* & o* -> cmxs & so"
-  ~prods:["%.cmxs"; x_dll]
-  ~dep:"%.mldylib"
-  ~doc:"Builds a .cmxs (native archive for dynamic linking) containing exactly \
-        the modules listed in the corresponding .mldylib file."
-  (Ocaml_compiler.native_shared_library_link_mldylib "%.mldylib" "%.cmxs");;
-
 rule "ocaml: mllib & p.cmx* & p.o* -> p.cmxs & p.so"
   ~prods:["%.cmxs"; x_p_dll]
   ~dep:"%.mllib"
@@ -376,6 +364,18 @@ rule "ocaml: mllib & cmx* & o* -> cmxs & so"
         could be found."
   (Ocaml_compiler.native_shared_library_link_mldylib "%.mllib" "%.cmxs");;

+rule "ocaml: mldylib & p.cmx* & p.o* -> p.cmxs & p.so"
+  ~prods:["%.p.cmxs"; x_p_dll]
+  ~dep:"%.mldylib"
+  (Ocaml_compiler.native_profile_shared_library_link_mldylib "%.mldylib" "%.p.cmxs");;
+
+rule "ocaml: mldylib & cmx* & o* -> cmxs & so"
+  ~prods:["%.cmxs"; x_dll]
+  ~dep:"%.mldylib"
+  ~doc:"Builds a .cmxs (native archive for dynamic linking) containing exactly \
+        the modules listed in the corresponding .mldylib file."
+  (Ocaml_compiler.native_shared_library_link_mldylib "%.mldylib" "%.cmxs");;
+
 rule "ocaml: p.cmx & p.o -> p.cmxs & p.so"
   ~prods:["%.p.cmxs"; x_p_dll]
   ~deps:["%.p.cmx"; x_p_o]

```

Changes
```diff
@@ -26,6 +26,9 @@ NEXT_RELEASE:
 - #127: install ocamlbuild's man pages, missing since 4.02
   (Gabriel Scherer)
 
+- #132: add a rule producing .cmxs from .mllib files
+  (Armaël Guéneau, requested by Daniel Bünzli)
```
gasche: Could you add just a bit more information here for people that have not followed the discussion? Something like "(Previously `.mllib` would only produce `.cm{x,}a`, and you had to use another file `.mldylib` for `.cmxs`. Now, a `.mllib` can also produce a `.cmxs`, but `.mldylib` files still take precedence when they exist.)"

Armael: Done (I think).

gasche:
Now I'm thinking of something else that we could test: if the `.mllib` file has the same name as a `.cmx` file, which of the two will take precedence to build a `.cmxs`? We want the `.mllib` file to take precedence, I suppose. Could you add a test for this? (Having a test for the `.mldylib` case would also make sense.)

dbuenzli:
That was quick, thanks Armael !

Armael:
Added in https://github.com/ocaml/ocamlbuild/commit/69ef32cf7c77ec71bf2dbafb871d9534f72c42bb

gasche:
(It seems that there is a complication, see the discussion in #131 again.)

PR #133, opened 2016-12-13
Remove warning 3.

bschommer:
This removes the warning 3 from ocamlbuild by using the correct upper/lowercase functions. Also the warn on error for all warnings is added. All places were the uppercase/lowercase functions were used were either for internal use or for module names which only are allowed to use a-z,A-Z,0-9 and _.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ OCAMLLEX  ?= ocamllex
 endif
 
 CP        ?= cp
-COMPFLAGS ?= -w L -w R -w Z -warn-error +a -I src -I +unix -safe-string -bin-annot
+COMPFLAGS ?= -w L -w R -w Z -I src -I +unix -safe-string -bin-annot
 LINKFLAGS ?= -I +unix -I src
 
 PACK_CMO= $(addprefix src/,\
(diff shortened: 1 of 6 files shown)


gasche:
I don't want to have `-warn-error +a` because it means that the code may stop building with new OCaml versions that introduce new warnings. One could try to argue that not having the flag means we fail to heed to advice of important warnings, but the fact that the only warning currently raised by OCamlbuild is the rather unimportant (and justifiably¹ delayed) `_ascii` transition would suggest otherwise.

¹: Using `foo_ascii` means that the code is incompatible with OCaml versions (< 4.03). It would not have been a good idea during the transition period around the 4.03 release where we would sometimes build separate-ocamlbuild from older OCaml installations. I think it is fine to have it now, as we can fairly assume that newer ocamlbuild releases will only be built from OCaml distributions with no integrated ocamlbuid, that is (>= 4.03).

bschommer:
I understand that it can be problematic to enable warnings as errors, so I remove -warn-error again from the compile options.

gasche:
(Squashed and) merge, thanks!

PR #134, opened 2016-12-22
Precise scale in verbose option description

Octachron:
This short commit precises the scale of the verbosity level in the corresponding option description.
The exposed scale {0,…,8} (aka {2,…,10} once mapped to log level)  is a strict subset of the real scale {-4,…,8} used by the logger. However, I thought that exposing the negative part of the scale (which has non-trivial interaction with other options) was not worth the added complexity.


gasche:
Thanks!

PR #137, opened 2016-12-26
Create installation directory for man pages.

atsampson:


gasche:
Thanks!

gasche:
@atsampson, does the build without this fix fail on your system? (What is your system?). I would like to understand if many users would be affected -- if I should make a bugfix release with just this, given that I just released 0.10.0.

As I'm looking into this, I'm wondering if it is not an issue with the non-opam default setting for `INSTALL_MANDIR`: I used `$(PREFIX)/man`, but on systems that install OCaml in `/usr/lib/ocaml` (and thus have `/usr/lib` as $(PREFIX), this gives `/usr/man/man1` instead of the expected `/usr/share/man/man1`. In this case, the "right" fix is not so much to create the file but to use the correct path.

gasche:
I changed the default manual installation location in #138.

atsampson:
@gasche, it fails for me because I'm installing into a completely empty destination directory while building a Stow package, so changing the MANDIR default wouldn't fix it -- you have to make sure you're creating all the directories you need explicitly (as you already do for the other directories).

However, PREFIX/share/man is the (Linux FHS) standard place to install man pages these days, so that's definitely a sensible change to make anyway.

I don't think it's worth doing a bugfix release just for this, since it's easy to work around.

gasche:
Thanks for the extra informations. The past release didn't go as well as planned because there is a regression on `.cmxs` files (everytime one hopes that a potential regression will not occur in the wild, one is disappointed), so there will probably be another release soon anyway.

PR #139, opened 2016-12-26
`.cmxs` fix

gasche:
This PR explores one possible way out of the regression of #131.


dbuenzli:
I don't think this will work if we don't replicate the C link flags that were recorded in the `cmxa`, except those for the stubs.

gasche:
Clearly simplistic. Out.

PR #140, opened 2016-12-27
use `.mllib -> .cmxs` instead of `.cmxa -> .cmxs`

gasche:
I'm continuing to explore the "simple" solutions to #131, after #139 that was unconvincing.

This one replaces the rule `.cmxa -> .cmxs`, which makes it difficult to preserve the stubs and command flags, with a `.mllib -> .cmxs` as originally proposed by @dbuenzli and @Armael. The hope is that the general flag-gathering machinery, and the existing `<foo.cmx>: ...` tag lines in the user projects, will magically work out together to get exactly what is needed in the dynamic library. In presence of C stubs, it seems a bit too simple, see @dbuenzli's [feature wish on this](https://github.com/ocaml/ocamlbuild/issues/131#issuecomment-269249151).


whitequark:
The cmxs fix has been merged in master.

PR #159, opened 2017-02-26
Fix linking of .cmxs targets that link stubs.

whitequark:
See the individual commit messages. In general, this addresses the reasons [0.10.1 got reverted](https://github.com/ocaml/opam-repository/pull/8157), and also fixes several glaring omissions in the build system and test suite that made writing the test for the fix quite obnoxious.

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -393,12 +393,12 @@ rule "ocaml: cmx & o -> cmxa & a"
 
 rule "ocaml: p.cmxa & p.a -> p.cmxs & p.so"
   ~prods:["%.p.cmxs"; x_p_dll]
-  ~deps:["%.p.cmxa"; x_p_a]
+  ~deps:["%.p.cmxa"]
   (Ocaml_compiler.native_shared_library_link ~tags:["profile";"linkall"] "%.p.cmxa" "%.p.cmxs");;
 
-rule "ocaml: cmxa & a -> cmxs & so"
+rule "ocaml: cmxa -> cmxs & so"
   ~prods:["%.cmxs"; x_dll]
-  ~deps:["%.cmxa"; x_a]
+  ~deps:["%.cmxa"]
   ~doc:"This rule allows to build a .cmxs from a .cmxa, to avoid having \
         to duplicate a .mllib file into a .mldylib."
   (Ocaml_compiler.native_shared_library_link ~tags:["linkall"] "%.cmxa" "%.cmxs");;
--- a/src/ocaml_compiler.ml
+++ b/src/ocaml_compiler.ml
@@ -30,6 +30,11 @@ let forpack_flags arg tags =
     Ocaml_arch.forpack_flags_of_pathname arg
   else N
 
+let include_flags_for_deps deps =
+  let dirnames = List.union [] (List.map Pathname.dirname deps) in
+  let include_flags = List.fold_right ocaml_add_include_flag dirnames [] in
+  S include_flags
+
 let ocamlc_c tags arg out =
   let tags = tags++"ocaml"++"byte" in
   Cmd (S [!Options.ocamlc; A"-c"; T(tags++"compile");
@@ -37,7 +42,7 @@ let ocamlc_c tags arg out =
 
 let ocamlc_link flag tags deps out =
   Cmd (S [!Options.ocamlc; flag; T tags;
-          atomize_paths deps; A"-o"; Px out])
+          include_flags_for_deps deps; atomize_paths deps; A"-o"; Px out])
 
 let ocamlc_link_lib = ocamlc_link (A"-a")
 let ocamlc_link_prog = ocamlc_link N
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -143,13 +143,22 @@ src/ocamlbuildlib.cmxa: src/ocamlbuild_pack.cmx $(EXTRA_CMX)
 
 # The packs
 
+# Build artifacts are first placed into tmp/ to avoid a race condition
+# described in https://caml.inria.fr/mantis/view.php?id=4991.
```
gasche: Would it not be simpler to just have `src/ocamlbuild_pack.cmx` depend on `src/ocamlbuild_pack.cmi`? This corresponds to serializing (byte before native), but it makes for simpler code (and makes sure there are no side-effects related to running in a different directory).

src/ocaml_specific.ml
```diff
@@ -393,12 +393,12 @@ rule "ocaml: cmx & o -> cmxa & a"
...
 rule "ocaml: p.cmxa & p.a -> p.cmxs & p.so"
   ~prods:["%.p.cmxs"; x_p_dll]
-  ~deps:["%.p.cmxa"; x_p_a]
+  ~deps:["%.p.cmxa"]
   (Ocaml_compiler.native_shared_library_link ~tags:["profile";"linkall"] "%.p.cmxa" "%.p.cmxs");;
 
 rule "ocaml: cmxa & a -> cmxs & so"
   ~prods:["%.cmxs"; x_dll]
-  ~deps:["%.cmxa"; x_a]
+  ~deps:["%.cmxa"]
```
gasche: if `.a` is not a dependency, the rule text/name should be `cmxa -> cmxs & so`.

Makefile
whitequark: @gasche It doesn't work. The ocamlopt command touches the cmi file, leading to a spurious rebuild afterwards.

gasche: Ah... Now we understand https://github.com/ocaml/ocamlbuild/pull/128!

src/ocaml_specific.ml
whitequark: Updated.

gasche:
I would still like to do a release without any of the `.cmxs` features included, as was my original idea, to make sure that the safe features proposed by the users before it get their foot in the released door. I am fine with doing also a release with the feature-related parts of this PR right after that.

whitequark:
@gasche CI is now green here.

whitequark:
@gasche So, how exactly (in terms of workflow) do you imagine making this intermediate release?

gasche:
Two things:

1. With the previous `.cmxs` / `.mllib` (wrong) changes, I could convince myself that the common ocamlbuild users that don't use `.cmxs` without a `.ml(dy)lib` would not see any behavioral change -- in particular, no regression. Your proposed patch touches a codepath that absolutely everyone uses. How do we convince ourselves that there are no regressions? (I suppose that the reasoning is: the linked functions are only called at "link" time, and for normal link-time usage `-I` flags mostly don't matter, the list of cases where it does matter is X, Y and Z, and for all of these the change improves the situation.)

2. My guts still tell me that it would be wise to do a release *without* any of the too-ambitious cmxs/mllib changes (including the toolchain feature for example), by reverting the currently broken commits, (we could call this 10.x or 11.0) and then a second release with those changes (11.0 or 12.0). Does that sound acceptable to you? I won't have time to work on this tonight, so it would only be later this week, which means I may feel bad about delaying the release of your work even more.

whitequark:
> I suppose that the reasoning is: the linked functions are only called at "link" time, and for normal link-time usage -I flags mostly don't matter, the list of cases where it does matter is X, Y and Z, and for all of these the change improves the situation.

Yes, this is exactly my argument.

How about just installing a non-insignificant amount of packages using ocamlbuild?

gasche:
> So, how exactly (in terms of workflow) do you imagine making this intermediate release?

My idea was:

1. Revert the cmxs/mllib changes
2. change the Changelog to present things as if the broken 10.x release had not happened, except a brief mention that they contained regressions and that we will just forget about them. Write the current Changes (the new features that were contributed since 9.x) in the section for this "safe" release (we need to decide whether to call it 10.x or 11.0; no strong opinion here)
3. Apply the cmxs/mllib again thanks to your fix.
4. Do another release (bumping the major version number).

We can merge the present PR before all of this (it would be reverted, and then un-reverted), or only at step (3).

Ideally I would have liked to test an opam build with the post-merge state (with the cmxs/mllib features in) *before* releasing it on the opam-repository, but I don't have the energy to setup this experiment right now.

whitequark:
> My guts still tell me that it would be wise to do a release without any of the too-ambitious cmxs/mllib changes (including the toolchain feature for example)

Is this really any better than the already-released 0.9.3? That's a lot of effort to expend on a number of minor fixes, *and* we still have to do the exactly same amount of testing for any more ambitious release on top of that.

gasche:
This is the reasoning I had when deciding to skip this step and release 10.x, and then I regretted it.

That said, I am not in a strong position for negotiation right now. If you strongly think that we should go ahead and just release 11.x, I am willing to be convinced. (Especially if I have the impression impression that you might help mitigate potential regressions since 0.9.x that would appear after the release.)

whitequark:
> That said, I am not in a strong position for negotiation right now. If you strongly think that we should go ahead and just release 11.x, I am willing to be convinced. (Especially if I have the impression impression that you might help mitigate potential regressions since 0.9.x that would appear after the release.)

Yes, that's the idea, on both counts. In fact if anyone pinged me before reverting 0.10.x we might have avoided this whole thing.

gasche:
Ok, let's go with it then. I propose to merge this and open a separate issue to discuss release coordination -- I can do some Changelog stuff, but I think it would be nice if *you* sent the announce when it happens. Do you have plans on the testing front?

whitequark:
> Ok, let's go with it then. I propose to merge this and open a separate issue to discuss release coordination -- I can do some Changelog stuff, but I think it would be nice if you sent the announce when it happens.

Sounds good.

> Do you have plans on the testing front?

Yes, I'm running some tests now.

PR #172, opened 2017-02-27
Test ocamlbuild on Travis CI

whitequark:


whitequark:
@gasche Any objections here? I've redone the OCAMLBUILDLIB change to be much less fragile, both internally and externally (see 0c24b7d).

gasche:
No objection at all in principle. I had the impression that I saw a preliminary version of this PR with only the `.travis.yml` file, without the preliminary commits, and I wanted to convince myself that it actually worked without the other change. I may have just dreamed, or you may have rebased, but this looks good. (I have not tested the Makefile/testsuite changes, but I trust you that they work.)

gasche:
(There is this mesmerizing "Some checks haven't completed yet" that makes me want to wait, but feel free to merge now, or I'll do it when I see the CI result.)

gasche:
So the test fails, of course, given that mtime fails. @whitequark is it your intention to have a failing CI as long as your fixing PR is not merged or the change reverted, or should we temporarily disable the `mtime` part? I am fine with either.

whitequark:
> I may have just dreamed, or you may have rebased

You've seen that. I rebase a lot, in general. Keeps the history clean.

> is it your intention to have a failing CI as long as your fixing PR is not merged

That is my intention indeed, because master is broken right now. No reason to hide that status.

PR #173, opened 2017-02-27
Test our external dependencies on Travis as well

whitequark:
The commit 48a3e14f is causing a spurious failure because of the way
opam sets up ocamlfind path. The opam PPA uses OCaml 4.02, which
installs a system camlp4. This system camlp4 is not afterwards used
by any opam package, but it is still present in the ocamlfind path,
which causes the subsequent error:
  Fatal error: OCaml and preprocessor have incompatible versions

It seems futile to try and fix this error in ocamlbuild testsuite
when simply installing our test dependencies will also do that,
and also get us more tests, so do that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,10 @@ install:
   - sudo apt-get install -qq opam
   - export OPAMYES=1
   - opam init --compiler=${OCAML_VERSION}
-  - opam install ocamlfind
+  - opam install ocamlfind camlp4 menhir
+  # The command above has ocamlbuild as a transitive build dependency.
+  # Remove it now, so as not to accidentally link to the system ocamlbuildlib.
+  - opam remove ocamlbuild
 script:
   - eval `opam config env`
   # First, run our own testsuite. Do this *before* any ocamlbuild is installed globally,


gasche:
I find `opam remove ocamlbuild` a bit fragile (we rely on the build-time-ness of the dependencies), there might be other way to ensure that we don't accidentaly use system stuff. But the change is better than what was before, so we may as well merge.

whitequark:
> I find opam remove ocamlbuild a bit fragile (we rely on the build-time-ness of the dependencies), there might be other way to ensure that we don't accidentaly use system stuff.

There's `/usr/bin/ocamlbuild` too and it was very easy to accidentally use it before 92d90f4d9de80890809b0853354b9c260e0b83bc. But yes.

whitequark:
@gasche Um, you were actually right about the fragility of it. In fact removing ocamlbuild removed camlp4 and menhir, and by some inexplicable mechanism this has caused the offending camlp4 test to pass.

whitequark:
@gasche I think I'll just remove the `opam remove ocamlbuild` part in a subsequent PR, after verifying the build log manually. I am sufficiently convinced that 48a3e14 made the testsuite not too fragile to do this.

whitequark:
#174

PR #174, opened 2017-02-27
Do not try to avoid preinstalled ocamlbuild on Travis

whitequark:
This didn't actually work, because it caused camlp4 and menhir
to be subsequently removed. The change to the testsuite in
the commit 92d90f4d should be robust enough that we can afford this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ install:
   - export OPAMYES=1
   - opam init --compiler=${OCAML_VERSION}
   - opam install ocamlfind camlp4 menhir
-  # The command above has ocamlbuild as a transitive build dependency.
-  # Remove it now, so as not to accidentally link to the system ocamlbuildlib.
-  - opam remove ocamlbuild
 script:
   - eval `opam config env`
   # First, run our own testsuite. Do this *before* any ocamlbuild is installed globally,


gasche:
Thanks.

(Another interesting piece of testing infrastructure is the bootstrapping logic in the `bootstrap/` directory. If you were to feel extra-brave, it would be interesting to look at integrating it in the CI infrastructure. I apologize for not having done this dirty but long-term essential work before; this has been on my plate since the split, but I never got around to it.)

whitequark:
> (Another interesting piece of testing infrastructure is the bootstrapping logic in the bootstrap/ directory. If you were to feel extra-brave, it would be interesting to look at integrating it in the CI infrastructure. I apologize for not having done this dirty but long-term essential work before; this has been on my plate since the split, but I never got around to it.)

Oh, is that logic still alive? I thought it was vestigial, like ocamlbuildlight. I admit I'm not sure how much of it makes sense, given the added pain of bootstrapping. The only positive outcome I see here is the lack of necessity of having a shell on Windows. But... the compiler itself will require a shell for foreseeable future, so what's the point?

gasche:
I moved the bootstrapping logic to `bootstrap`, it is never used in the common build/install path, for testing purposes only. Testing that ocamlbuild is able to build itself is an interesting exercise, I think, more realistic than the `testsuite` tests.

whitequark:
https://github.com/ocaml/ocamlbuild/pull/175 (the CI will fail because I didn't include the cmxs fix in that branch).

PR #177, opened 2017-02-28
Test more downstream code on CI

whitequark:


gasche:
While we are at it, could we have an `ocb_stubblr`-using dependency? It would make sense to have one dependency for all non-trivial ocamlbuild plugins out there: cppo also comes to mind; there is also the namespace-as-ocamlbuild plugin, but I don't know if it is used in opam-released packages, and Solvuu (again, not sure).

whitequark:
> While we are at it, could we have an ocb_stubblr-using dependency? It would make sense to have one dependency for all non-trivial ocamlbuild plugins out there: cppo also comes to mind

Added both.

> Solvuu (again, not sure).

The dependencies seem quite heavy (they all need core_kernel...)

gasche:
Ok, this sounds like a nice compromise. Feel free to merge once we get CI results or whenever.

gasche:
(Are you the kind of person that prefer other people to merge their PRs, or to have a chance to do a last amending round before merging themselves?)

whitequark:
> (Are you the kind of person that prefer other people to merge their PRs, or to have a chance to do a last amending round before merging themselves?)

It does not matter to me very much. Unless explicitly noted my PR commit stacks are always ready to merge; I rebase before every push.

Issue #176, opened 2017-02-27
11.0 release preparation


gasche:
@whitequark this is a thread to discuss the 11.x release

What I have to do before the release is to reorder the Changelog to keep a trace of 10.x, yet clarify to users that all the features since 9.x are, practically speaking, new in 11.x. 

I wonder what we should do (and have the time/energy to do) on the testing front. The CI is already helpful (tahnks!), and obviously if we break OPAM we will know pretty soon, but there is no easy way for me to test a development version before the release.

dbuenzli:
> but there is no easy way for me to test a development version before the release.

Is it still not possible to reliably pin `ocamlbuild` ? I can certainly try to live on a pin for the next few days and check my `opam reinstall $(opam list -s --installed --depends-on ocamlbuild)` install base behave correctly. Just give me a reliable tag or commit to pin to.

gasche:
Yes, `opam pin add ocamlbuild .` should work correctly. (I don't remember about this not being possible in the past.)

whitequark:
> Is it still not possible to reliably pin ocamlbuild ?

It is perfectly possible and in fact our CI relies on that.

whitequark:
@gasche I have found one instance of an affected downstream package. Ironically, my own, ppx_deriving. I am not calling this a "regression" because it is merely exposing a bug that already was there; I would be inclined to call the *previous* behavior a bug.

Look here: https://github.com/whitequark/ppx_deriving/commit/c2fb119f#diff-7c9f76fe069008cd250e1fb989cb43ff. Before 0.11.0, this .mllib file got ignored. Now it is not, and ppx_deriving_make.cma correctly packs only ppx_deriving_create.cma...

whitequark:
This is possibly related to https://github.com/ocaml/ocamlbuild/pull/140, https://github.com/ocaml/ocamlbuild/issues/166 and https://github.com/ocaml/ocamlbuild/issues/42...

whitequark:
@gasche BTW, since the CI is green now: I went ahead and, for master, disabled force pushes, and added a requirement that the build pass before a PR can be merged.

whitequark:
@gasche  I've now installed, with --build-test, a fairly large amount of packages:
```
# Installed packages for 4.04.0+ocamlbuildtest:
alcotest                          0.7.2  Alcotest is a lightweight and colourful test framework.
astring                           0.8.3  Alternative String module for OCaml
[...]

PR #178, opened 2017-02-28
Changes fix in preparation to 11.0

gasche:
See #176.

### DIFF
--- a/Changes
+++ b/Changes
@@ -11,34 +11,16 @@ We also use PR#$N for pull requests.
 Changes breaking compatibility are marked with a "* " bullet instead
 of the usual "- ".
 
+
 NEXT_RELEASE:
 -------------
 
-- "noautolink" tag for ocaml{c,opt}
-  (Gabriel Scherer)
-
-0.10.1 (26 Dec 2016):
----------------------
-
-- #124, #136: do not explicitly pass -shared when building shared libraries;
-  let the compiler decide what to build.
-  (whitequark)
-
-- #137, #138: fix the man/ installation for non-opam users
-  (Adam Sampson and Gabriel Scherer)
-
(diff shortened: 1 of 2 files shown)


Changes
```diff
@@ -14,31 +14,12 @@ of the usual "- ".
...
-OCamlbuild 0.10.0 introduces a minor change to the way `.cmxs` files
-are produced when no `.mldylib` file is absent: it will now use the
-exact same semantics as `.cmxa` and `.mllib` file -- in particular, it
+OCamlbuild 0.11.0 introduces a change to the way `.cmxs` files are
+produced when no `.mldylib` file is absent: it will now use the exact
+same semantics as `.cmxa` and `.mllib` file -- in particular, it
 should not be necessary anymore to have identical
 `foo.{mllib,mldylib}` files, only `foo.mllib` should suffice. See the
-detailed Changelog for details.
+detailed Changelog below for details.
```
whitequark: Nit: the file is called "Changes" (was it not?), so this should probably be "detailed change log".

gasche: Thanks, fixed. (I kept the attached "changelog" substantive.)

gasche:
@whitequark, the CI on this PR seems to never terminate, while it is reported as successful on https://travis-ci.org/ocaml/ocamlbuild/builds/206057002 . Would you have any idea of what is happening? Maybe if Travis is unreliable, we should disable the "merge button disallowed while required CI runs are not complete" option?

whitequark:
> Would you have any idea of what is happening?

Yes. Unfortunately, Amazon S3 is down in US-EAST-1, which caused internet-wide outage of any infrastructure that relies on it. This happens once every few years.

It's [recovering](status.aws.amazon.com) now, so let's just wait a bit.

gasche:
Ok, the Changes will wait. Sorry for the noise.

PR #227, opened 2017-03-05
also install signatures.{cmi,cmti}

### LINKED ISSUE #217: MPR#5243: Improve the API documentation
_PR transferred from https://caml.inria.fr/mantis/view.php?id=5243_
[original reporter: @Chris00]

Ideally, the API documentation for plugins should be more complete (basically serve as a reference how to write a plugin) and cross-referenced.
http://nicolaspouillard.fr/ocamlbuild/html/Signatures.PLUGIN.html

gasche:
Hopefully fixes #217.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,7 +235,7 @@ clean::
 
 # The library is put in LIBDIR/ocamlbuild. We copy
 # - the META file (for ocamlfind)
-# - src/signatures.mli (user documentation)
+# - src/signatures.{mli,cmi,cmti} (user documentation)
 # - the files in INSTALL_LIB and INSTALL_LIB_OPT (if available)
 
 # We support three installation methods:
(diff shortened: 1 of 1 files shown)


gasche:
In fact I was a bit too optimistic about the usability of those languages. OCaml pull-requests (that pull from ocamlbuild master) are now failing with:

```
cp META src/signatures.{mli,cmi,cmti} /home/travis/local/lib/ocamlbuild
cp: cannot stat `src/signatures.{mli,cmi,cmti}': No such file or directory
make[1]: *** [install-lib-basics] Error 1
```

Oh well. Will fix.

whitequark:
@gasche How did this happen?

dbuenzli:
Mmmh sorry about the merge then. It's strange the CI was saying yes. Is due to a lack `bin_annot` ?

gasche:
I'm just as confused (no need to apologize!); `-bin-annot` is passed correctly (see the full [failing logs](https://travis-ci.org/ocaml/ocaml/jobs/208169496)), I wonder if this is just a case of the `foo.{bar,baz}` pattern not being expanded -- but I don't understand why that would be, might this be a bashism that some other shells don't support? I certainly cannot reproduce any install/uninstall issue on my machine.

dbuenzli:
> I wonder if this is just a case of the foo.{bar,baz} pattern not being expanded -- but I don't understand why that would be, might this be a bashism

Actually when I saw it I wanted to comment that it is a bashism but then thaught that this might be supported by the `make` language. So yes, it is a bashism.

PR #231, opened 2017-03-14
Provide `-project` command line option to avoid recursive traversal hint

hannesm:
The current heuristic (Options.ocamlbuild_project_heuristic) checks for either
`_tags` or `myocamlbuild.ml` to be present, or "-recursive" being passed to
ocamlbuild.  This works fine for many applications, but e.g. for mirage
unikernels it does not (they are in a flat directory, subdirectories may be
present containing data which should not be traversed).

At the moment, mirage creates an empty `myocamlbuild.ml` before calling
ocamlbuild to avoid this (for MirageOS users misleading) hint.

(Suggestions for the command line option are welcome.)  The bits of mirage in question are at https://github.com/mirage/mirage/blob/279f365d8a031ccd30cb593657396f916d09a4ba/lib/mirage.ml#L1617-L1627

### DIFF
--- a/src/ocaml_compiler.ml
+++ b/src/ocaml_compiler.ml
@@ -124,7 +124,7 @@ let prepare_compile build ml =
         dprintf 3
           "Warning: Failed to build the module %s requested by ocamldep."
           name;
-        if not (!Options.recursive || Options.ocamlbuild_project_heuristic ())
+        if not (!Options.recursive || !Options.project || Options.ocamlbuild_project_heuristic ())
         then Log.at_failure ~name:"a module failed to build,
            while recursive traversal was disabled by fragile heuristic;
            hint that having a _tags or myocamlbuild.ml would maybe solve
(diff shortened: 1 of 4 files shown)


gasche:
I'm not fond of `project` as an option name, because to me it suggests way more than a choice of recursive traversal or not. I would propose the following: to have `-recursive` as a long option for `-r`, and `-non-recursive` for its negation. If either of the two options are explicitly set, the user expressed their intent, and there is no need for the "project" heuristic and corresponding warning.

Unfortunately, `Options.recursive : bool ref` is part of the OPTIONS signature in signatures.mli; changing it to `bool option ref` as would be natural risks breaking user plugins. I have thought about schemes were you pass an `recursion_explicitly_set` flag on the side, but that is fragile (with respect to users setting `Options.recursive` directly), so I think the best thing to do is to make a breaking change and use `bool option`.

hannesm:
closing -- time has shown that this does not have a high enough priority for me, and in addition to that, [my use case is soon to-be-migrated away from ocamlbuild](https://github.com/mirage/mirage/issues/969)

PR #233, opened 2017-03-19
Don't hard-code object file extension...

dra27:
https://github.com/ocaml/ocamlbuild/pull/172/commits/691ccdc355fd1a887fffdfbf518b29ee70e8eaf6#diff-b67911656ef5d18c4ae36cb6741b7965R161 in #172 breaks MSVC. One-line fix...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,7 +163,7 @@ src/ocamlbuild_pack.cmx: $(PACK_CMX)
 	mkdir -p tmp
 	$(OCAMLOPT) -pack $^ -o tmp/ocamlbuild_pack.cmx
 	mv tmp/ocamlbuild_pack.cmx src/ocamlbuild_pack.cmx
-	mv tmp/ocamlbuild_pack.o src/ocamlbuild_pack.o
+	mv tmp/ocamlbuild_pack$(EXT_OBJ) src/ocamlbuild_pack$(EXT_OBJ)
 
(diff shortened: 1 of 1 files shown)


gasche:
Do I correctly understand that 0.11.0 is broken under MSVC? Should we get a 0.11.1 out? Try to setup so Appveyor CI to catch similar issues in the future?

dra27:
It is broken - though the fix is very obvious and any MSVC user will be building from sources, I'm sure! fdopen already patched his repository 6 days ago - I hit this during my OPAM rebase. So up to you how urgent 0.11.1 is - I probably wouldn't.

AppVeyor will slow your PRs down a lot, because OCaml is slow to build (relative to Unix) - but I was thinking the OCaml's AppVeyor should be testing camlp4 and ocamlbuild, as Travis does.

whitequark:
> AppVeyor will slow your PRs down a lot

It can't be longer than one hour since that's the Appveyor limit, and I think one hour is fine, so why not?

dra27:
Sure, if you're happy with that! It's just a lot slower than your Travis check. I may be able to do something towards it this week, if that's of any use (as I want to set-up AppVeyor for another project next week)

whitequark:
Sounds good. I was going to set up Appveyor but could not discover the motivation for that.

PR #234, opened 2017-04-13
Add flags afl_instrument and afl_inst_ratio.

yallop:
See https://github.com/ocaml/ocaml/pull/504.

### DIFF
--- a/Changes
+++ b/Changes
@@ -93,6 +93,10 @@ detailed changelog below for details.
 - "noautolink" tag for ocaml{c,opt}
   (Gabriel Scherer)
 
+- #234: add "afl_instrument" and "afl_inst_ratio" flags for corresponding
+  ocamlopt options.
+  (Jeremy Yallop)
+
 0.10.{0,1} (Dec 2016):
 ----------------------
 
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -788,6 +788,8 @@ let () =
     (fun param -> S [A "-inline-max-unroll"; A param]);
   pflag ["ocaml"; "native"; "compile"] "unbox_closures_factor" ~doc_param:"10"
     (fun param -> S [A "-unbox-closures-factor"; A param]);
+  pflag ["ocaml"; "native"; "compile"] "afl_inst_ratio" ~doc_param:"50"
+    (fun param -> S [A "-afl-inst-ratio"; A param]);
   ()
 
 let camlp4_flags camlp4s =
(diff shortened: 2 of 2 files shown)


whitequark:
@yallop Could you add a changelog entry?

gasche:
Now that ocamlbuild versions are dissociated from ocaml versions, a delicate problem arises: how should we handle OCaml features that are only supported in some of the versions that ocamlbuild support? This is, of course, not specific to this PR, but as far as I know this is the first PR raising this issue.

We could either have enough logic in the ocamlbuild rules to not add the flags on unsupported OCaml versions (that may make more sense in some cases than others; for `-bin-annot` I think that silently ignoring the flag would be ok, but for `-afl-fuzz` it is not as clear), or aggressively update the required OCaml versions for ocamlbuild whenever we add support for new OCaml compiler flags. Or a mix of the two. What do we think?

whitequark:
> or aggressively update the required OCaml versions for ocamlbuild whenever we add support for new OCaml compiler flags.

This basically leaves out dry everyone on older OCaml versions with respect to bugfixes. For example, we've recently introduced a compile time regression with `.mldylib` files. If we bump the version requirement now, everyone using 4.04 or earlier will get stuck with a slower ocamlbuild (or, if they install an even older version, with a buggier ocamlbuild...)

> We could either have enough logic in the ocamlbuild rules to not add the flags on unsupported OCaml versions (that may make more sense in some cases than others; for -bin-annot I think that silently ignoring the flag would be ok, but for -afl-fuzz it is not as clear)

It seems unlikely that afl_fuzz will be used in any way but a carefully controlled one during a QA process, so it might not even be necessary to gate it in any way. I propose deferring this decision until we are up the wall against it :-)

gasche:
> This basically leaves out dry everyone on older OCaml versions with respect to bugfixes. 

We *could* consider maintaining bugfix branches.

>  I propose deferring this decision until we are up the wall against it :-)

Ok.

I have no other comment on the PR and would be in favor of merging once we have the changelog entry.

whitequark:
> We could consider maintaining bugfix branches.

I am strongly against that. It is a large amount of work, and a larger than necessary amount of difficult decisions on whether this work should be done for a certain compiler version. Adding cppo annotations, while certainly ugly (and introducing a circular dependency), clearly has much less cumulative overhead.

yallop:
> @yallop Could you add a changelog entry?

Done: see e234958.

gasche:
I don't think we need preprocessing in this case, because we can express the conditional test in pure OCaml code (this is not about using language features that older compiler would barf on); see [this in Batteries' myocamlbuild.ml](https://github.com/ocaml-batteries-team/batteries-included/blob/4f10969/myocamlbuild.ml#L219-L223) for an example:

```
      if ocaml_version = (4, 0) then begin
        (* OCaml 4.00 has -bin-annot but no ocamlbuild flag *)
        flag ["ocaml"; "bin_annot"; "compile"] (A "-bin-annot");
        flag ["ocaml"; "bin_annot"; "pack"] (A "-bin-annot");
end;
```

(This particular test is precisely on 4.00 because it implements a flag that exists in ocamlbuild 4.01 and above for a feature introduced in 4.00. In the ocamlbuild codebase itself those would be comparison rather than equality tests.)

whitequark:
Oh! I have not realized that we have `ocaml_version`; but indeed even if we did not, it would be simpler to add it than to drag in cppo.

yallop:
I just realised that I added the changelog entry to the wrong section -- sorry!  Fixed in #235.

PR #239, opened 2017-06-07
examples: add example of working with dependent projects

henrytill:
I found the following example on `ocaml.org` useful when I needed a quick and easy way include code from another local subproject without the added overhead of using an `opam`-based workflow:
https://ocaml.org/learn/tutorials/ocamlbuild/Working_on_dependent_projects_with_ocamlbuild.html

This PR adds a slightly simpler example of the same technique to the `examples` directory , using the `04-library/libdemo` modules as the library code on which the new example project depends.  It also amends the _Examples_ section of the manual to include the new example.

### DIFF
--- a/examples/07-dependent-projects/Makefile
+++ b/examples/07-dependent-projects/Makefile
@@ -1,0 +1,33 @@
+#
+# Pure OCaml, project depends on library code from a local project
+#
+# For more details, see:
+# https://ocaml.org/learn/tutorials/ocamlbuild/Working_on_dependent_projects_with_ocamlbuild.html
+#
+
+# - The -I flag introduces sub-directories
+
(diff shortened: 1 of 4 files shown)


gasche:
This looks fine, thanks! One nitpick: I think it would be nice to point back to the [original page](https://ocaml.org/learn/tutorials/ocamlbuild/Working_on_dependent_projects_with_ocamlbuild.html), both in a comment in the example Makefile and in the description in the manual.

henrytill:
Great - thanks for the review.  I think your suggestion is a good one.  I'll update and re-push.

henrytill:
Updated.

gasche:
Thanks! Merged.

Issue #69, opened 2016-04-12
OCAMLLIB overrides `ocamlbuild -where`, which is no longer correct in OPAM environment


camlspotter:
Now OCamlBuild can be installed by OPAM and its library directory, typically `$HOME/.opam/<switch>/lib/ocamlbuild` is different from `$OCAMLLIB/ocamlbuild` when the "system" compiler is used.  However, `ocamlbuild_where.ml` still checks the variable and override the lib dir when it is defined.

camlspotter:
OASIS also assumes ocamlbuild library directory is `ocamlc -where`/ocamlbuild. See: https://forge.ocamlcore.org/tracker/index.php?func=detail&aid=1538&group_id=54&atid=291

camlspotter:
A workaround is not to use "system" compiler, but it is hard to switch in Cygwin OPAM + MinGW OCaml environment. (I think...)

gasche:
Should we just remove the lookup to OCAMLLIB, or maybe replace it by OCAMLBUILD_LIBDIR, which is the variable name we use at configuration time?

whitequark:
> Should we just remove the lookup to OCAMLLIB, or maybe replace it by OCAMLBUILD_LIBDIR, which is the variable name we use at configuration time?

Context: https://caml.inria.fr/mantis/view.php?id=4379. Do you think it should be removed? In my opinion, since the advent of package managers, it is very rare to want to use an ocamlbuild with an compiler other than the one it was built with, and such cases are sufficiently accomodated with -install-lib-dir, so I think it should be just removed and not replace with another environment variable.

In any case, the current behavior is incorrect.

whitequark:
On reflection, my reasoning is incorrect. Anyone trying to distribute binary builds of ocamlbuild (for Windows or elsewhere) will encounter this problem. But the environment variable solution is still quite bad, as it is not easy to diagnose and results in pollution of environment.

I wonder how ocamlfind copes with this?

dra27:
Just been bitten by this - ocamlbuild reading OCAMLLIB is not what caused me problems, it's the fact that it's not taken into account that it was installed using findlib (so it's `$OCAMLLIB/site-lib/ocamlbuild`). Even if you want the pain of trying to distribute binary builds for Windows, it would seem to be a good idea for ocamlbuild to fix the `/ocamllib` suffix at configure-time based on how ocamlbuild is configured to be installed.

gasche:
[...]

PR #240, opened 2017-06-08
Add new heuristic for handling OCAMLLIB

dra27:
This is a cut at implementing the idea in #69. I've lazily done it dynamically on src/ocamlbuild_where.ml rather than working out a boolean flag at build time, but I don't think this matters.

### DIFF
--- a/src/ocamlbuild_where.ml
+++ b/src/ocamlbuild_where.ml
@@ -11,10 +11,40 @@
 (*                                                                     *)
 (***********************************************************************)
 
-let bindir = ref Ocamlbuild_config.bindir;;
+module O = Ocamlbuild_config;;
+
+let bindir = ref O.bindir;;
 let libdir = ref begin
-  Filename.concat
-    (try Sys.getenv "OCAMLLIB"
-     with Not_found -> Ocamlbuild_config.libdir)
-    "ocamlbuild"
+  let root, suffix =
+    let ocaml_lib_len = String.length O.ocaml_libdir + 1 in
+    let lib_len = String.length O.libdir_abs in
+    (* Windows note: O.ocaml_libdir and O.libdir_abs have both been passed
+       through GNU make's abspath function and will be forward-slash normalised.
+       Filename.dir_sep is therefore not appropriate here. *)
+    if lib_len < ocaml_lib_len
+       || String.sub O.libdir_abs 0 ocaml_lib_len <> O.ocaml_libdir ^ "/" then
+      O.libdir, "ocamlbuild"
+    else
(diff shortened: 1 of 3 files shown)


src/ocamlbuild_where.ml
```diff
@@ -11,10 +11,27 @@
...
+  let root, suffix =
+    let ocaml_lib_len = String.length O.ocaml_libdir + 1 in
+    let lib_len = String.length O.libdir_abs in
+    if lib_len < ocaml_lib_len then
+      O.libdir, "ocamlbuild"
+    else
+      let prefix =
+        String.sub O.libdir_abs 0 ocaml_lib_len
+      in
+      if prefix = O.ocaml_libdir ^ "/" then
```
gasche: Wouldn't it be more robust to use `O.ocaml_libdir ^ Filename.dir_sep`, and compute the length of this for the `String.sub` call?

gasche: Also: I'd rather have the short case first and the long case as an `else` (so `prefix <> ...` I guess).

src/ocamlbuild_where.ml
```diff
@@ -11,10 +11,27 @@
...
+        try
+          let root =
```
gasche: calling this `root` is a bit confusing. `subroot`?

src/ocamlbuild_where.ml
dra27: Blimey, did I hard-code a forward slash... using Ubuntu for a day is clearly some kind of disease :blush:

dra27:
It turns out I was using `"/"` rather than `Filename.dir_sep` for a reason - I have therefore commented it! Code updated to rename root to subroot as suggested and invert the branches (in fact, I merged the two `if`s)

src/ocamlbuild_where.ml
```diff
@@ -11,10 +11,40 @@
...
+      try
+        let normalise_slashes =
+          if Sys.os_type = "Win32" then
+            String.map (function '/' -> '\\' | c -> c)
+          else
+            function s -> s
+        in
+        let subroot =
+          String.sub O.libdir_abs ocaml_lib_len (lib_len - ocaml_lib_len) |>
+          normalise_slashes
```
gasche: My personal taste (well, of course I believe it is the only reasonable choise) is to have the infix operator at the start of the next line, `|> normalize_slashes`, but in any case I am happy with code readability now.

dra27: Yeah :blush:

dra27:
Pipe tidied up and Changes entry added

PR #253, opened 2017-07-30
Only run native tests if ocamlopt is available

infinity0:
Otherwise the tests fail on architectures without ocamlopt, e.g. https://buildd.debian.org/status/fetch.php?pkg=ocamlbuild&arch=mips64el&ver=0.11.0-1&stamp=1501347472&raw=0

Reading Makefile.config to check ocamlopt existence, is not the most elegant; I'd be happy to change if there's a better way.

### DIFF
--- a/testsuite/internal_test_header.ml
+++ b/testsuite/internal_test_header.ml
@@ -1,8 +1,16 @@
 #load "unix.cma";;
 
+#mod_use "../src/ocamlbuild_config.ml";;
+
 #use "ocamlbuild_test.ml";;
 
 module M = Match;;
 module T = Tree;;
 
 let _build = M.d "_build";;
+
+let ocamlopt_available =
+  if Ocamlbuild_config.ocaml_native then
+    Fullfilled
+  else
+    Missing ("ocamlopt")
--- a/Changes
+++ b/Changes
@@ -14,6 +14,9 @@ of the usual "- ".
 NEXT_RELEASE:
 -------------
 
+- #253: only run native tests if ocamlopt is available.
+  (Ximin Luo, review by whitequark and Gabriel Scherer)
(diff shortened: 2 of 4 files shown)


testsuite/internal_test_header.ml
```diff
@@ -6,3 +6,21 @@ module M = Match;;
 module T = Tree;;
 
 let _build = M.d "_build";;
+
+let ocamlopt_available =
+  let chan = open_in "../Makefile.config" in
+  let rec loop () =
+    try
+      let line = String.trim (input_line chan) in
+      let len = String.length line in
```
whitequark: This can all be replaced with `line = "OCAML_NATIVE=true"`, can't it?

gasche:
Another way to do this, which I personally think would be nicer, would be to call `#mod_use "../src/ocamlbuil_config.ml";;` to get configuration values.
But then there is the issue that OCAML_NATIVE is not currently exported in this config file, so you would need to add a line for this in `configure.make`. I think that having two extra declarations, `ocaml_native` and `ocaml_native_tools`, would be fine -- long-term it would be nice to reorder `ocamlbuild_config` to follow the more regular `Makefile.config` structure, but that is for another PR.

infinity0:
Ah yes that is nicer, I did not notice `src/ocamlbuild_config.ml` before. I've updated the commit.

gasche:
Very nice, thank you. Before I merge, would you like to add a change entry in the `Changes` file for the next release? I think the credit line "(Ximin Luo, review by whitequark and Gabriel Scherer)" would be fine.

infinity0:
`Changes` updated.

gasche:
Thanks! Merged.

Issue #190, opened 2017-03-01
MPR#6103: Add ocamlbuild support for ocamlfind's -ppopt option


damiendoligez:
_PR transferred from https://caml.inria.fr/mantis/view.php?id=6103_
[original reporter: @gasche]

The current ocamlbuild support for `-pp` flags is not satisfying for preprocessing commands that need user-passed arguments. It is possible to pass `-pp foo` and `-ppflag foo` to the command-line invocation of ocamlbuild, but they will replace the preprocessor invocation command instead of being appended to it.

Consider the following file test.ml:
```
  IFDEF FOO THEN
    print_endline "foo"
  ELSE
    print_endline "bar"
  ENDIF
```
The following will work:
```
  ocamlbuild -tag "pp(camlp4of -DFOO)" test.byte
```
the following will also work:
```
  ocamlbuild -ppflag camlp4of -ppflag -DFOO test.byte
```
they both generate the command
```
  .../ocamlc.opt -c -pp 'camlp4of -DFOO' -o test.cmo test.ml # cached
```

but the following will fail:
```
  ocamlbuild -use-ocamlfind -tag "syntax(camlp4o)" -package camlp4.macro -ppflag -DFOO test.byte

  + /usr/bin/ocamlfind ocamldep -package camlp4.macro -syntax camlp4o -pp -DFOO -modules test.ml > test.ml.depends
  Warning: -pp overrides the effect of -syntax partly
  sh: 0: Illegal option -D
  Preprocessing error on file test.ml
```

It would be useful to directly support ocamlfind's `-ppopt` option
(when `-use-ocamlfind` is used), that doesn't replace the preprocessor
command entirely but only appends to it. One would then be able to do
(I'm using below the new `-syntax` command-line option corresponding
to the flag):
```
  ocamlbuild -use-ocamlfind -syntax camlp4o -package camlp4.macro -ppopt -DFOO test.byte
```

Of course, it is important to also have a `ppopt(-DFOO)` parametrized
flag to be able to use it locally.

Note that, as a workaround, it is currently possible to pass `-ppopt` as
a `-cflags` parameter:
```
  ocamlbuild -use-ocamlfind -syntax camlp4o -package camlp4.macro -cflags -ppopt,-DFOO test.byte
```

The problem is that only works during "compile" steps, and will not
work, for example, when building a `.pp.ml` target.

[...]

PR #265, opened 2017-12-30
add ppopt(arg) and ppxopt(package,arg) when -use-ocamlfind

### LINKED ISSUE #45: ppopt and ppxopt options
It doesn't seem possible to add these ocamlfind options in the `_tag` file

gasche:
This is a pull request to fix #45, #190. I haven't thought very hard about the feature, and this is a minimalist implementation, that only passes the options unchanged to `ocamlfind`, and in particular is only available when `-use-ocamlfind` is used.

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -693,13 +693,19 @@ let () =
       ["c"; "compile"];
     ] in
 
-    (* tags package(X), predicate(X) and syntax(X) *)
+    (* tags package(X), predicate(X) and syntax(X), ppopt(X), ppxopt(X) *)
     List.iter begin fun tags ->
       pflag tags "package" ~doc_param (fun pkg -> S [A "-package"; A pkg]);
       pflag tags "predicate" ~doc_param:"archive"
         (fun pkg -> S [A "-predicates"; A pkg]);
-      if List.mem "ocaml" tags then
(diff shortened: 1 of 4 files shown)


src/ocaml_specific.ml
```diff
@@ -693,13 +693,19 @@ let () =
...
     List.iter begin fun tags ->
       pflag tags "package" ~doc_param (fun pkg -> S [A "-package"; A pkg]);
       pflag tags "predicate" ~doc_param:"archive"
         (fun pkg -> S [A "-predicates"; A pkg]);
-      if List.mem "ocaml" tags then
-        pflag tags "syntax" ~doc_param:"camlp4o" (fun pkg -> S [A "-syntax"; A pkg])
+      if List.mem "ocaml" tags then begin
+        pflag tags "syntax" ~doc_param:"camlp4o"
+          (fun pkg -> S [A "-syntax"; A pkg]);
+        pflag tags "ppopt" ~doc_param:"camlp4-arg"
```
whitequark: -pp is not just camlp4, isn't it? ocaml-m17n also uses -pp.

gasche: I reused the name from the `ocamlfind` documentation, and it is just informative (there is no semantics on the `doc_param`). Yet your point is completely correct and I will change it to use `pparg` instead.

Issue #268, opened 2018-03-12
How to generate only a statically linked clib?


hannesm:
I want to build only a statically linked c library, as done by passing `custom` or `failsafe` to `ocamlmklib`, but I fail to find a way how to pass such an option via `_tags` (and/or `myocamlbuild.ml`) to the build system.

repository in question is [mirage-xen](https://github.com/mirage/mirage-xen) on the `topkg` branch.  Any hints?

hannesm:
In the latest commit on the branch mentioned above, the target is only `libmirage-xen_bindings.a`, I don't understand why ocamlbuild (or ocamlmklib) nevertheless tries to build the shared object.

gasche:
I actually know little about static vs. dynamic linking in the non-OCaml world, so I am not sure how to test whether my experiments work or not. Please be patient :-) Also, I haven't had the time to build all of mirage-xen's dependecies, so I decided to experiment with a small repro case.

I have a file `lib.ml` and `lib.mli` with dummy content. The only rule in `ocamlbuild -documentation` that mentions `ocamlmklib` is the `.clib` one, so I created a `libfoo.clib` file with the following content:

```
lib.o
```

then `ocamlbuild -verbose 1 -tag custom libfoo.a` (or listing `true: custom` in _tags) runs the following commands:

```shell
ocamldep.opt -modules lib.ml > lib.ml.depends
ocamldep.opt -modules lib.mli > lib.mli.depends
ocamlc.opt -c -o lib.cmi lib.mli
ocamlopt.opt -c -o lib.cmx lib.ml
ocamlmklib.opt -o foo lib.o
```

If I understand correctly, the lack of a `-custom` flag passed to `ocamlmklib` in the final call is the issue for you. This should clearly be fixed upstream, but in the following I wrote the following `myocamlbuild.ml` script:

```ocaml
open Ocamlbuild_plugin

let _ = dispatch begin function
  | After_rules ->
    flag ["ocamlmklib"; "custom"] & A"-custom";
  | _ -> ()    
end
```

With this plugin file, I get the following behavior for the same command:

```shell
ocamldep.opt -modules lib.ml > lib.ml.depends
ocamldep.opt -modules lib.mli > lib.mli.depends
ocamlc.opt -c -o lib.cmi lib.mli
ocamlopt.opt -c -o lib.cmx lib.ml
ocamlmklib.opt -o foo -custom lib.o
```

(the `-custom` flag is indeed passed.)

[...]

PR #269, opened 2018-03-12
#268: add flag support for ocamlmklib options

gasche:
See #268

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -884,6 +884,11 @@ flag ["ocaml"; "link"; "library"; "custom"; "byte"] (A "-custom");;
 flag ["ocaml"; "link"; "toplevel"; "custom"; "byte"] (A "-custom");;
 flag ["ocaml"; "compile"; "profile"; "native"] (A "-p");;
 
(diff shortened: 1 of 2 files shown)


src/ocaml_specific.ml
```diff
@@ -1020,4 +1025,15 @@ begin
...
 end;;
 
+begin
+  let ocamlmklib_pflag ?doc_param flag =
+    pflag ["ocamlmklib"] flag ?doc_param
+      (fun param -> S [A ("-"^flag); A param]) in
+  ocamlmklib_pflag "cclib";
+  ocamlmklib_pflag "ccopt";
+  ocamlmklib_pflag "dllpath";
+  ocamlmklib_pflag "rpath";
```
whitequark: I don't think we need both `-dllpath` and `-rpath` since those are aliases. I would say keep only `-rpath`.

gasche: Thanks for the comment. My idea was that windows users may find "dllpath" more natural, and unix users prefer "rpath". I'll move to rpath-only.

whitequark: Windows doesn't have the concept of rpath.

gasche:
Thanks!

PR #273, opened 2018-03-26
add the "link" tag to "pack" actions (fixes #272)

### LINKED ISSUE #272: the `pack` action must also pass `package(...)` options to ocamlfind
A [build failure](https://github.com/mirleft/ocaml-nocrypto/issues/143) of nocrypto comes down to the fact that ocamlbuild does not forward `-package` setting in `pack` actions, which somehow(?) results in the typer creating subtly-incompatible inferred interfaces for the generated `.cmi` file. See the diagnosis by @diml in https://github.com/ocaml/opam-repository/pull/11628#issuecomment-375697444.

This is an ocamlbuild bug and I'm creating this issue to track a fix.

gasche:
Instead of a separate category, "pack" is now another form of linking
like "program", "library", "toplevel" and "output_obj". This fixes the
issue that package(...) tags where not passed at pack-production time,
spotted by Jérémie Dimino in
  <https://github.com/ocaml/opam-repository/pull/11628#issuecomment-375697444>
  <https://github.com/janestreet/ppx_sexp_conv/issues/20>
More generally, this extends the meaning of all "link" flags to
"pack", which seems to be the correct behavior for all the rules we
inspected.

I tested that this change fixes the original 'nocrypto' issue
(once their _tags is also fixed to provide the right package(..)
option at pack-time).

This is a somewhat invasive change to the behavior of "pack" and
"link" tags, which may affect all user-defined flag declarations. An
alternative to this change would be to just add a specific "package"
rule when the "pack" tag is present, but my understanding is that
linking options quite generally apply to "pack" productions
(that really behave mostly like library-archive production already
under the ("link";"library") regime), so this more invasive change is
more correct: the extra cases where it applies were likely to be small
bugs before.

(I checked that ocb-stubblr is orthogonal to this change and does not
need any update.)

### DIFF
--- a/Changes
+++ b/Changes
@@ -22,6 +22,15 @@ NEXT_RELEASE:
   custom, debug, failsafe, linkall, ccopt(..), cclib(..), rpath(..), ldopt(..)
   (Gabriel Scherer, report by Hannes Mehnert, review by whitequark)
 
+- #272: add the "link" tag to "pack" actions
+  Instead of a separate category, "pack" is now another form of linking
+  like "program", "library", "toplevel" and "output_obj". This fixes
+  the issue that package(...) tags where not passed at pack-production
+  time, spotted by Jérémie Dimino. More generally, this extends
+  the meaning of all "link" flags to "pack", which seems to be the correct
+  behavior for all the rules we inspected.
+  (Gabriel Scherer, original issue diagnosis by Jérémie Dimino)
+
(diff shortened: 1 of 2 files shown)


gasche:
I decided to go ahead and merge this PR which fixes a known issue. I'm thinking of doing a new ocamlbuild release.

Issue #152, opened 2017-02-24
MPR#6292: ocamlbuild -I does not understand paths that aren't relative descendants


damiendoligez:
_PR transferred from https://caml.inria.fr/mantis/view.php?id=6292_
[original reporter: @dsheets]

When running ocamlbuild with the `-I` option to include a directory for compilation unit reference, if the provided directory is not a relative descendant of the current working directory, then an error is produced e.g.:

```
Failure: Included or excluded directories must be implicit (not "../lib").
```

This restriction is probably no longer necessary and could be removed.

mtzguido:
Hi, I also ran into this. It is also failing for me if I provide fully explicit paths, even if they are a subdirectory of the current one. For instance `ocamlbuild -I .` might work but `ocamlbuild -I $(pwd)` fails.

Any chance of changing this?

gasche:
I gave an opinion on this mid-2018 in https://github.com/ocaml/ocamlbuild/pull/279#issuecomment-392315358 ; the fundamental reason why ocamlbuild insists on relative paths is that a path given by the user may be reinterpreted either at the invocation site, or within `_build`.

The only way to accept explicit paths that would make sense is to force them to only point to subdirectories of `$(pwd)`, and the trim the `$(pwd)` prefix. I personally find this semantics a bit surprising / irregular, and I'm not sure it's a good idea to add such "magic" to the tool itself. (Given the moribund state of ocamlbuild usage and the ever-so-exciting of the rest of my work, I prefer to focus on development issues that are more critical.)

Note that it can be done rather easily from the caller side, at least if using `bash`: instead of `ocamlbuild -I foo`, just use `path=foo; ocamlbuild -I ${foo#$(pwd)/}`, which strips the `$(pwd)/` prefix from the variable.

PR #279, opened 2018-05-25
Remove two checks of Pathname.is_implicit

charlesetc:
This is an attempt to address issue #152. I wanted to share this code and bump #152 in general. It'd be great to allow passing absolute paths to `-I` in ocamlbuild.

I have very little familiarity with ocamlbuild's source and absolutely no confidence that these are the right changes to make. These changes do allow a local ocamlbuild to build with relative include paths on my computer, and all the tests pass :paintbrush:

### DIFF
--- a/src/options.ml
+++ b/src/options.ml
@@ -383,17 +383,10 @@ let init () =
   reorder show_tags show_tags_internal;
   reorder plugin_tags plugin_tags_internal;
 
-  let check_dir dir =
-    if Filename.is_implicit dir then
-      sys_file_exists dir
-    else
-      failwith
-        (sprintf "Included or excluded directories must be implicit (not %S)" dir)
-  in
   let dir_reorder my dir =
     let d = !dir in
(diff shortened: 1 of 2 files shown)


charlesetc:
I think these changes are buggy when using absolute paths, so I'm closing this PR. I hope absolute paths are possible someday

gasche:
I hadn't got around to looking in depth at your code yet, apologies. (It did look a bit surprising on a first skim.)

I'm not terribly optimistic about ways to extend ocamlbuild to support absolute paths, because of the way we rely on relative paths being re-interpreted in a different way in `_build/` (compared to the project root).

I think that if your workflow would strongly benefit from a change here, you might be better off evaluating alternative build systems. In particular, dune/jbuilder, who is actively developed and should succeed to ocamlbuild, could be a reasonable option.

charlesetc:
Ah well I guess that makes sense — thanks very much anyways!

PR #282, opened 2018-08-13
Fix compilation for OCaml 4.08

xclerc:
The trunk version of OCaml has recently introduced
a function named `Format.dprintf`, but `ocamlbuild`
already had such a function -with a different signature-
in the `Log` module.

This PR tweaks the order of `open` statements so that
the function from `Format` does not shadow the one
from `Log`.

### DIFF
--- a/src/main.ml
+++ b/src/main.ml
@@ -14,12 +14,12 @@
 
 (* Original author: Berke Durak *)
 open My_std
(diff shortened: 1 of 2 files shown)


gasche:
Thanks! I would like to do a release soon.

PR #283, opened 2018-08-22
Resolved conflict between Log.dprintf and Format.dprintf

nandor:
Ocaml introduces [Format.dprintf](https://github.com/ocaml/ocaml/pull/1959) which conflicts with Log.dprintf defined in ocamlbuild.

This PR ensures the correct dprintf is called.

### DIFF
--- a/src/main.ml
+++ b/src/main.ml
@@ -14,7 +14,6 @@
 
 (* Original author: Berke Durak *)
 open My_std
-open Log
 open Pathname.Operators
 open Command
 open Tools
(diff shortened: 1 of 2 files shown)


gasche:
Thanks for the patch! Unfortunately, I believe that there may be a duplication of efforts here, as it is very similar to the (slightly simpler) #282 by @xclerc. Can you confirm that #282 solves your issue?

nandor:
Yes it does, I am closing this issue.

PR #285, opened 2018-09-09
Test latest revisions of compilers

dra27:


dra27:
Travis is running on PRs? See https://travis-ci.org/ocaml/ocamlbuild/builds/426309983 for example?

gasche:
... that must have been a brain fart on my side, sorry. Thanks for the PR!

dra27:
Actually, especially on mobile, I think the some relatively recent GUI changes have made passing checks less obvious, so perhaps not so much a brainfart! On mobile, it’s just the green tick which gets me to the Travis status

PR #289, opened 2018-09-21
Build with -strict-sequence

Armael:
This is a small fix that allows ocamlbuild to build when -strict-sequence is enabled.

I found this while trying to investigate the impact of enabling -strict-sequence by default (see https://github.com/ocaml/ocaml/pull/1971) -- ocamlbuild breaks, which transitively breaks a lot of packages.

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,3 +1,5 @@
 S src
 S bootstrap
 B .
+FLG -strict-sequence
+FLG -safe-string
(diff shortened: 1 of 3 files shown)


gasche:
I'm not sure this is the right fix, I would rather restrict `pp_elt` to return `unit`.

Armael:
That was my original idea, but I didn't want to (potentially) break client code by changing the API (even though I agree this is quite unlikely).

Armael:
OK. I amended the PR.

gasche:
Merged, thanks!

PR #293, opened 2018-12-14
change divisor and dividend

chakku000:
In the example of 05-lex-yacc, when the DIV is used, the divisor and dividend was not correct.  The divisor is used as the dividend and the dividend is used as divisor.
So, `./main.native 1/2` outputs `2`, and `./main.native 2/1` outputs `0`.

And, I change the divisor and dividend.


gasche:
Indeed! Thanks, I have no idea where the mistake came from.

PR #298, opened 2019-06-23
Change link to examples in CONTRIBUTING

Lereena:
The past link led to 404. As far as I understood from the context, the directory in which the examples lie was meant, not the manual section, so I changed it to a link to the actual directory.


gasche:
Thanks! Merged.

Issue #300, opened 2019-08-26
Race condition in Shell.mkdir_p


mroch:
https://github.com/ocaml/ocamlbuild/blob/4ef4b18223383f79819ae123f6dff2a7b969fa7a/src/shell.ml#L62-L64

If another process creates the directory after the `sys_file_exists` call but before the `mkdir`, then `mkdir` raises.

This has been happening in the wild for Flow (https://github.com/facebook/flow/issues/8027). I suspect it's because of `make -j` and we have multiple `ocamlbuild` targets so they run in parallel.

A safer approach could be to suppress the `EEXIST`: https://github.com/janestreet/core/blob/e8cf8a6d340c4af1e8177591782fb3e326eb0a0b/src/core_unix.ml#L1463-L1478

gasche:
Thanks for the report.

A very safe thing to be doing would be to replace `; mkdir dir` by `; try_mkdir dir` -- check again for the file existence after the recursive parent creation. This is, of course, not removing the race entirely, but it is an easy change to reason about. The actually-correct approach you suggest sounds good, but harder to implement (we would have to cross the abstraction boundaries of ocamlbuild's system-utils abstractions, and I don't understand the portability consequences of such a change).

It's actually easy to reproduce this failure, using a makefile that looks as follows:

```makefile
DIR=foo
FILE=code.ml

setup:
	mkdir -p $(DIR)
	echo "let x = 1" > $(DIR)/$(FILE)
	echo "run 'make -j all' to reproduce the failure'"

clean:
	ocamlbuild -clean

dist-clean: clean
	rm $(DIR)/$(FILE)
	rmdir $(DIR)

.PHONY: all test1 test2
all: test1 test2

test1:
	$(MAKE) _build/$(DIR)/$(FILE)

test2:
	$(MAKE) _build/$(DIR)/$(FILE)

_build/$(DIR)/$(FILE):
	ocamlbuild $(DIR)/$(FILE)
```

I will first test the safe-and-incomplete change; if it fixes the reproduction case, maybe that's a good first step and I can do a minor release. Then I will look at the harder-and-complete fix.

gasche:
So:
- the simple fix does not in fact fix the race
- a more complex "simple fix" is to just use `mkdir -p` on Unix systems, which does fix the race by delegating concurrency control to the filesystem (I like that approach); this requires Windows-specific logic (Windows' `mkdir` does not support the `-p` flag), for now I kept the legacy implementation on Windows.
[...]

PR #302, opened 2019-08-31
Fix race conditions when invoking `ocamlbuild` in parallel from a makefile

gasche:
I'm creating this PR to track my progress on fixing #300 and related race-conditions issue. The current reproduction testcase is as follows:

```
DIR=foo
FILE=code.ml

setup:
	mkdir -p $(DIR)
	echo "let x = 1" > $(DIR)/$(FILE)
	echo "run 'make -j all' to reproduce the failure'"

clean:
	ocamlbuild -clean

dist-clean: clean
	rm $(DIR)/$(FILE)
	rmdir $(DIR)

.PHONY: all test1 test2
all: test1 test2

test1:
	$(MAKE) _build/$(DIR)/$(FILE)

test2:
	$(MAKE) _build/$(DIR)/$(FILE)

_build/$(DIR)/$(FILE):
	ocamlbuild $(DIR)/$(FILE)
```

running `make setup; make -j all` exposes races in ocamlbuild's parallel accesses to the filesystem.

I don't think we will ever try to guarantee that concurrent ocamlbuild invocations work robustly, but we should at least fix the issues that users encounter in the wild, when reasonably possible.

### DIFF
--- a/ocamlbuild.opam
+++ b/ocamlbuild.opam
@@ -7,7 +7,7 @@ license: "LGPL-2 with OCaml linking exception"
 doc: "https://github.com/ocaml/ocamlbuild/blob/master/manual/manual.adoc"
 dev-repo: "git+https://github.com/ocaml/ocamlbuild.git"
 synopsis:
-  "OCamlbuild is a build system with builtin rules to easily build most OCaml projects."
+  "OCamlbuild is a build system with builtin rules to easily build most OCaml projects"
 
 build: [
   [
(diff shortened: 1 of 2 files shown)


gasche:
I tried to deal with a race in `mkdir_p` and another in `rm_f`, and now the reproduction script works reliably on my machine. I'm thinking of merging this, getting a minor release out, and waiting for more user reports to do more if needed.

Issue #301, opened 2019-08-30
build ocamlbuild fail


samikshro:
Hi! I am trying to install ocamlbuild on my computer (OS X High Sierra 10.13.6), but it keeps failing. I have attached the error message below. I installed opam using Homebrew, and my compiler version is 4.08.1.

Here are the commands I ran before I got this issue:
opam init --bare -a -y
opam switch create 4.08.1 ocaml-base-compiler.4.08.1
eval $(opam env)
opam install -y utop ounit qtest yojson lwt lwt_ppx menhir ansiterminal lambda-term merlin ocp-indent user-setup bisect_ppx-ocamlbuild

<img width="1280" alt="Screen Shot 2019-08-29 at 11 13 59 AM" src="https://user-images.githubusercontent.com/26678068/64047049-aa2b8a80-cb3b-11e9-9358-0a82ca70b004.png">

gasche:
I have no idea what is going wrong here, and I cannot reproduce with 4.08.1+0.14.0 on my Linux machine.

Could you access a more complete error log? The display of the help message suggests that a command-line was passed to the compiler that it does not recognize, so maybe a version mismatch, but I don't see how that would happen.

Another thing that would help would be to see the content of the `Makefile.config` file generated during the build (to check that the paths refer to the right opam switch, with the right compiler version). Could you try this? The easiest way to do would be to get the sources (by cloning the git repository for example; building from `master` is fine) and use `opam install . --inplace-build`, which will build in place and thus create `Makefile.config` in the current directory.

dbuenzli:
Just in case try again after an `opam udpate` that sometimes helps. 

To find the `Makefile.config` @gasche mention rather use `opam install --keep-build-dir` and simply go hunt for the `Makefile.config` in the build directory mentioned in `path` field of the error message. Also post the contents of the file mentioned in the `output-file` field of the error message somewhere.

samikshro:
Just cloned the git repository and generated the Makefile.config. Here it is!
# This file was generated from configure.make

OCAML_PREFIX=
OCAML_BINDIR=/Users/samikshro/.opam/4.08.1/bin
OCAML_LIBDIR=/Users/samikshro/.opam/4.08.1/lib/ocaml
OCAML_MANDIR=/Users/samikshro/.opam/4.08.1/man

EXT_OBJ=.o
EXT_ASM=.s
EXT_LIB=.a
EXT_DLL=.so
EXE=

OCAML_NATIVE=true
[...]

PR #303, opened 2019-09-05
rename user-rebindable {LINK,COMP}FLAGS into OCB_{LINK,COMP}FLAGS

gasche:
This should help avoid build failures due to unintential
redefined-variable capture, see #301.

(failure report by Samik Shrotriya)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,8 @@ OCAMLLEX  ?= ocamllex
 endif
 
 CP        ?= cp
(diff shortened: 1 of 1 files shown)


dbuenzli:
Except for the lack of a changes entry this looks good (also I suspect you did a `git grep` of these variables and replaced all of them, I didn't check this)

hhugo:
@gasche, I think we should merge this one

Issue #304, opened 2019-12-04
cmxs rules are mentioning .so without producing them


ivg:
All cmxs producing rules are mentioning a `.so`  as a product, though it is not able to produce it, at least on a Linux machine. The `link_shared` rule is creating a call to `ocamlopt` with the `-shared` option, which doesn't produce `.so` nor it is compatible with the `output-obj` option. 

This issue has a significant impact on (re)build times as it prevents the cache system from working properly. The cache system looks for the so file and retriggers linking for each library every time, even when nothing is changed.

gasche:
Fixed by #305.

PR #305, opened 2019-12-04
removes .so from products of all cmxs rules

ivg:
These cmxs rules do not produce .so files as they promise. This confuses the cache system, see #304 

I don't know what effect it may have on other systems, like Windows or Mac, but if `ocamlopt -shared` produces cmxs on both it should work (and I believe, that this is what ocamlopt is doing).

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -365,13 +365,13 @@ rule "ocaml: mllib & cmx* & o* -> cmxa & a"
         see the rules producing a .cmxs from a .mllib or a .mldylib."
   (Ocaml_compiler.native_library_link_mllib "%.mllib" "%.cmxa");;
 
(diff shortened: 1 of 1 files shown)


gasche:
Your remark seems perfectly correct today. But I would prefer to understand why this `.so` was included in of the targets. Was this always a mistake, or did the behavior of the compiler change over time? Why? (Given that it took some care to use the system-appropriate extension, I'm skeptical about the first option.)

gasche:
The `dll` code was part of the initial cmxs-related commit: 8f2181f9fe14330e2cebcc1e5a82be71790fd64c.

gasche:
I can't observe any sort of `.so` production on 3.12.1, 4.01 or 4.02 either.

gasche:
Given these observations, I will be happy to merge the PR once the CI is happy.

ivg:
forgot to write the magic fixes #304

PR #311, opened 2022-01-06
Fix asciidoc error

jamesjer:
With asciidoc 9.1.0 and pygments 2.10.0, I get this error from running `asciidoc manual/manual.adoc`:
```
asciidoc: ERROR: manual.adoc: line 1364: undefined filter attribute in command: pygmentize -f html -l {language} {src_numbered?-O linenos=table} {encoding?-O encoding={encoding}} {args=}
```
and line 1364 does not appear in the resulting HTML file.  With this commit, there are no errors and line 1364 appears in the output.

### DIFF
--- a/manual/manual.adoc
+++ b/manual/manual.adoc
@@ -1359,7 +1359,7 @@ rule "ocaml C stubs: clib & (o|obj)* -> (a|lib) & (so|dll)"
 One cannot define a rule if there already exists a built-in rule of the same name.
 Directly including the rule `"ocaml dependencies ml"` used above in your plugin would result in a failure when loading the plugin:
 
-[source]
+[source,ocaml]
 ----
   Rule.Exit_rule_error("Rule.add_rule: already exists: (ocaml dependencies ml)").
 ----


gasche:
Thanks! This looks like a useful fix.

But: this is weird, it looks like your version of asciidoc does not support source blocks that do not indicate a specific language anymore? (Was it like this from the start, or a compatibility-breaking change that occurred recently?) Some [documentation material](https://docs.antora.org/antora/latest/asciidoc/source/) for AsciiDoc still advertises `[source]` as valid syntax.

I'll go ahead and merge (once the CI decides that it is happy), but probably not cut out a new release, given that this is a documentation-only issue -- this manual is not installed anyway, so having the fixed version in the git repository should be good enough.

jamesjer:
Well, if I run asciidoc like this:
```
asciidoc --attribute=language=ocaml manual/manual.adoc
```
then I don't get an error.  So, yes, it looks like my copy of asciidoc insists on knowing the language.  [This page](https://asciidoc-py.github.io/source-highlight-filter.html) seems to suggest that the language attribute is mandatory.  It looks like plain `[source]` is okay if a `:language:` attribute has been set previously.

PR #312, opened 2022-01-07
Fix testsuite on OCaml > 4.07

jamesjer:
Three of the internal tests fail on recent versions of OCaml due to changes in error output.  With this change, I got a successful run of the testsuite with OCaml 4.07.2, 4.08.2, 4.11.3, 4.12.2, and 4.13.1.


gasche:
Thanks! It would be nice to have a more robust test, but in the meantime your fix is sensible and will do.

PR #314, opened 2022-01-28
OCaml 5.0 compatibility

jonludlam:

PR #315, opened 2022-01-28
Add support for OCaml 5.00

kit-ty-kate:
Same as #314 but this also keeps backward compatibility with OCaml 4.03

### DIFF
--- a/src/main.ml
+++ b/src/main.ml
@@ -11,6 +11,7 @@
 (*                                                                     *)
 (***********************************************************************)
 
+(* Compatibility with both OCaml < 4.08 and >= 5.00 *)
 module Pervasives = struct
   let exit = exit
 end
--- a/src/main.ml
+++ b/src/main.ml
@@ -11,6 +11,9 @@
 (*                                                                     *)
 (***********************************************************************)
 
+module Pervasives = struct
+  let exit = exit
+end
 
 (* Original author: Berke Durak *)
 open My_std
@@ -20,8 +23,6 @@ open Command
 open Tools
 open Ocaml_specific
 ;;
-
-module Pervasives = Pervasives [@ocaml.warning "-3"]
 
 exception Exit_build_error of string
 exception Exit_silently
--- a/src/my_std.ml
+++ b/src/my_std.ml
@@ -11,11 +11,12 @@
 (*                                                                     *)
 (***********************************************************************)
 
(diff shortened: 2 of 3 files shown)


talex5:
Thanks - with this PR I was able to get ocamlbuild installed.

kit-ty-kate:
@gasche any chance this could be merged? I could do the release if you want to give me the rights if you don’t have the time.

src/main.ml
```diff
@@ -11,6 +11,9 @@
 (*                                                                     *)
 (***********************************************************************)
 
+module Pervasives = struct
+  let exit = exit
+end
```
gasche, line 16: If I understand, that's a very clever trick to write code that is compatible with both old OCaml versions (pre-Stdlib) and new OCaml versions. Right? If so, could you include a comment that says so? Because my first reaction was "let's just replace Pervasives.exit with Stdlib.exit!".

gasche (approved):
Sounds nice in practice. Thanks for the ping! I left a minor review comment. I will definitely give you release rights.

gasche:
A minor recollection from my memory (could be flawed). I think that you (@kit-ty-kate) were instrumental in pushing for ocamlbuild to become its own project, separate from the compiler distribution, and that at the time you argued that you would contribute to it if it was. Shortly after this discussion, we did split ocamlbuild from the compiler distribution, but at that point you were busy with other things and your sort-or-suggested contributions somehow didn't really materialize. It's fun that now you are the one keeping the lights on so that a few remaining OCaml projects can continue building as-is. (I think it's valuable effort.) Thanks a lot!

src/main.ml
```diff
@@ -11,6 +11,9 @@
 (*                                                                     *)
 (***********************************************************************)
```
kit-ty-kate, line 13:
```suggestion

(* Compatibility with both OCaml < 4.08 and >= 5.00 *)
```

src/my_std.ml
```diff
@@ -11,12 +11,13 @@
 (*                                                                     *)
 (***********************************************************************)
```
kit-ty-kate, line 13:
```suggestion

(* Compatibility with both OCaml < 4.08 and >= 5.00 *)
```

src/main.ml
kit-ty-kate: good point. Done

PR #317, opened 2022-04-16
Support new locations of unix, str and dynlink

dra27:
Updates ocamlbuild for compatibility with the changes proposed in ocaml/ocaml#11198 in order to silence the alert displayed by that PR.

### DIFF
--- a/src/ocaml_specific.ml
+++ b/src/ocaml_specific.ml
@@ -819,9 +819,19 @@ camlp4_flags' ["camlp4orr", S[A"camlp4of"; A"-parser"; A"reloaded"];
 
 flag ["ocaml"; "pp"; "camlp4:no_quot"] (A"-no_quot");;
 
-ocaml_lib ~extern:true "dynlink";;
-ocaml_lib ~extern:true "unix";;
-ocaml_lib ~extern:true "str";;
+let ocaml_otherlibs_lib =
+  match split_ocaml_version with
+  | Some (major, minor, _, _) when (major, minor) >= (5, 0) ->
+    fun name ->
+      let dir = "+" ^ name in
+      ocaml_lib ~extern:true ~dir name;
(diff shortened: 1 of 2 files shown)


src/ocaml_specific.ml
```diff
@@ -819,9 +819,19 @@ camlp4_flags' ["camlp4orr", S[A"camlp4of"; A"-parser"; A"reloaded"];
...
-ocaml_lib ~extern:true "str";;
+let ocaml_otherlibs_lib =
+  match split_ocaml_version with
+  | Some (major, minor, _, _) when (major, minor) >= (5, 0) ->
+    fun name ->
+      let dir = "+" ^ name in
+      ocaml_lib ~extern:true ~dir name;
+      flag ["ocaml"; "compile"; "use_" ^ name] (S[A"-I"; A dir])
+  | _ ->
+    fun name -> ocaml_lib ~extern:true name;;
```
gasche, line 830: Is `ocaml_otherlibs_lib "foo"` not something you can already do with `ocaml_lib ~extern:true ~dir:"+foo" "foo"`? I have the impression that the latter already adds `ocaml, compile, use_foo: -I +foo `.

Documentation:

https://github.com/ocaml/ocamlbuild/blob/e11720f866c0f4a794a9a8e2813a7352ab08d0c0/src/signatures.mli#L800-L815

Implementation:

https://github.com/ocaml/ocamlbuild/blob/e11720f866c0f4a794a9a8e2813a7352ab08d0c0/src/ocaml_utils.ml#L100-L134

gasche: (Maybe one needs `ocaml_lib ~extern:true ~dir:"+foo" "foo/foo"?)

hhugo:
This can be closed, fixed in #328

PR #319, opened 2022-09-27
Add support for OCaml 5.1

kit-ty-kate:
`ocamlbuildlib` and the `ocamlbuild` binary were wrongly compiled.

See https://github.com/dbuenzli/mtime/issues/45


gasche:
Thanks!

The problem that you are fixing is that the Makefile is incorrect, it compiles all source files with `-for-pack` but only some of them are packed. (I guess that before this mistake was ignored by the compiler, and now it isn't anymore.)

Your fix is to create new source directories: only the files to be packed remain in `src/`, other library files go in `plugin-lib/` and binaries go in `bin/`.

There are however two issues with the proposed change as-is, which should be relatively easy to fix:
- ocamlbuild is able to bootstrap itself (`make -C bootstrap`), this is useful for testing purposes, and I believe the change breaks this bootstrap; one would need to update the boostrap build to the new source locations
- as can be observed in the bootstrap build files in `bootstrap/`, which provide a rather clear specification of the intent of the ocamlbuild build, the first iteration of this PR puts some files in the wrong directory:
   + `ppcache` and `ocamlbuildlight` should go in `bin/`
   + the files remaining in `src/` should be only thosee listed in `bootstrap/ocamlbuild_pack.mlpack` or `Makefile/$PACK_CMO`

kit-ty-kate:
Done

gasche:
Very nice, thanks!

kit-ty-kate:
Done https://github.com/ocaml/opam-repository/pull/22182

gasche:
You're the best.

PR #324, opened 2023-09-19
Quote Makefile arguments to allow spaces

jonahbeckford:
This is the upstream of https://github.com/diskuv/diskuv-opam-repository/blob/main/packages/ocamlbuild/ocamlbuild.0.14.0/files/ocamlbuild-0.14.0.diskuvocaml2.patch

### DIFF
--- a/Changes
+++ b/Changes
@@ -14,6 +14,8 @@ of the usual "- ".
 NEXT_RELEASE:
 -------------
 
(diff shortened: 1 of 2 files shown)


gasche (approved):
This is obviously safe -- and actually not Windows-specific, unix users also may live the dangerous life and have spaces in their absolute paths. Thanks!

Issue #326, opened 2024-03-27
`ocamlbuild -where` is not relocatable


gridbugs:
`ocamlbuild -where` always returns the value of the `$OCAMLBUILD_LIBDIR` makefile variable set during build time. This means that if a pre-compiled instance of ocamlbuild is ever relocated, `ocamlbuild -where` will return the wrong result.

Practically, this causes problems for dune package management. Dune builds opam packages by running their build and install commands in a transient sandbox, after which all installed artifacts are moved into a shared opam-like filesystem and the sandbox is deleted. This results in errors like:
```
Warning: Won't be able to compile a native plugin
Failure: Cannot find "ocamlbuild.cmo" in ocamlbuild -where directory.
```
...because the directory returned by `ocamlbuild -where` was inside the (now deleted) build sandbox. (possibly related: https://github.com/ocaml/ocamlbuild/issues/321)

My plan for addressing this is to implement `ocamlbuild -where` by locating the library directory relative to the path to the `ocamlbuild` executable (by means of `Sys.executable_name`), possibly keeping the current behaviour as a fallback.

See also: https://github.com/ocaml/dune/issues/10290

gasche:
You report the following behavior on the dune issue:

```
$ _build/_private/default/.pkg/ocamlbuild/target/bin/ocamlbuild -where
/home/s/tmp/bonsai/_build/.sandbox/7158a9da761d39477e414a6f2dd9d3fc/_private/default/.pkg/ocamlbuild/target/lib/ocamlbuild
```

Two comments:

1. The standard way to deal with this issue in Linux distributions is to set the correct path at configure-time, that is, the path where the software will be *installed* and not the paths where the software is *built*. The configure logic of ocamlbuild does the following to set `OCAMLBUILD_LIBDIR`:

   ```
   OCAMLBUILD_LIBDIR ?= \
     $(or $(shell opam config var lib 2>/dev/null),\
          $(shell ocamlfind printconf destdir 2>/dev/null),\
          $(OCAML_LIBDIR))
   ```

   That is, it tries to ask standard tools about where ocamlbuild is going to be installed, and it lets the people invoking the script override the default logic by defining the OCAMLBUILD_LIBDIR variable yourself.

   Maybe the `dune` packaging script for ocamlbuild (where is it? can you point to the sources?) is not doing the right thing, by configuring ocamlbuild with a temporary sandbox prefix instead of using the final installation prefix?

[...]

PR #327, opened 2024-04-03
Attempt to guess correct libdir from current exe

gridbugs:
In cases where the libdir computed by `ocamlbuild -where` doesn't contain the expected ocamlbuild library files, attempt to guess an alternative libdir relative to the currently running executable's path (using `Sys.executable_name`). If the resulting libdir does contain the expected ocamlbuild library files then a warning is printed and the guessed libdir is used.

The motivation for this change is to allow the ocamlbuild (opam) package to be built by dune's package management features. Dune builds opam packages by executing their build and install commands in a temporary sandbox environment, and then copying the resulting artifacts to a final installation path. This creates a problem for ocamlbuild as the `OCAMLBUILD_LIBDIR` variable set at build time will be a path inside the temporary sandbox. That path is then used to compute the value printed by `ocamlbuild -where` which consequently reports a non-existant path.

This is an attempt to satisfy dune's constraint that a package's installed files must work after copying them out of their original install location.

We discussed this a little in https://github.com/ocaml/ocamlbuild/issues/326. This is intended to fix this issue: https://github.com/ocaml/dune/issues/10290.

### DIFF
--- a/src/ocamlbuild_where.ml
+++ b/src/ocamlbuild_where.ml
@@ -14,6 +14,19 @@
 module O = Ocamlbuild_config;;
 
 let bindir = ref O.bindir;;
+
+(* Check if a directory contains a file that libdir is expected to contain. *)
+let libdir_contains_ocamlbuild_library libdir =
+  Sys.file_exists (Filename.concat libdir "ocamlbuild.cma") ||
+    Sys.file_exists (Filename.concat libdir "ocamlbuild.cmx")
+
+(* Try to guess the libdir from the current exe's location. *)
+let guess_libdir_from_executable_name () =
+  let guessed_bin_dir = Filename.dirname Sys.executable_name in
+  Filename.concat
+    (Filename.concat (Filename.dirname guessed_bin_dir) "lib")
+    "ocamlbuild"
+
 let libdir = ref begin
   let root, suffix =
     let ocaml_lib_len = String.length O.ocaml_libdir + 1 in
@@ -46,5 +59,29 @@ let libdir = ref begin
         Sys.getenv "OCAMLLIB", Filename.concat subroot "ocamlbuild"
       with Not_found -> O.libdir, "ocamlbuild"
   in
-  Filename.concat root suffix
+  let libdir = Filename.concat root suffix in
+  if libdir_contains_ocamlbuild_library libdir then
+    libdir
+  else
+    (* The libdir doesn't contain the ocamlbuild library. Maybe the ocamlbuild
+       installation has been moved to a new location after installation. Try to
+       guess the libdir from the current exe's path. *)
+    let guessed_libdir = guess_libdir_from_executable_name () in
+    if libdir_contains_ocamlbuild_library guessed_libdir then (
+      Printf.eprintf "Warning: The library directory where ocamlbuild was \
+        originally installed (%s) either no longer exists, or does not contain \
+        a copy of the ocamlbuild library. Guessing that the correct library \
+        directory is %s because a copy of the ocamlbuild library exists at that \
+        location, and because it is located at the expected path relative to \
+        the current ocamlbuild executable. This can happen if ocamlbuild's \
+        files are moved from the location where they were originally \
+        installed."
+        libdir
+        guessed_libdir;
+      guessed_libdir)
+    else
+      (* The guessed path also doesn't contain the ocamlbuild library, so just
+         return the original libdir to help the user debug the error which will
+         likely result. *)
+      libdir
 end;;


gridbugs:
Closing this in favour of maintaining a fork of ocamlbuild in Dune's [opam-overlays](https://github.com/ocaml-dune/opam-overlays/tree/main/packages/ocamlbuild) repo. The dune developers intend to maintain forks of a small number of widely used packages whose built artifacts do not work correctly after relocation from dune's build sandbox.

kit-ty-kate:
Why close it though? Relocatable OCaml is coming anyway and it has nothing to do with dune. This PR only needed some regular pings for reviews (i personally don't have time right now but might in the future, but i'm not the only maintainer)

hhugo:
We could also make ocamlbuild depend on findlib and no longer hardcode theses paths

kit-ty-kate:
that sounds like a good idea to me

gridbugs:
Ok fair enough! I'm happy to leave it open but I don't have a lot of time to commit to working on this in the near future.

kit-ty-kate:
no problem!

gasche:
I have not done much on this PR because I find the proposed use-case odd, if not broken. The standard software configure-build-install is to configure software with information on where it is going to be installed, and then build and install it. Dune broke this convention in its internal package-manager implementation, by providing installation locations within its own sandbox. (@gridbugs explained this well in https://github.com/ocaml/dune/issues/10290#issuecomment-2024239930 ). I would naively think that trying to support this convention is the logical next step, but in fact dune maintainers decided to special-case packages that assume it; this closed-world approach is likely to make it difficult in the future to package software on opam-repository that use different build systems than `dune`, which I find worrying given `dune`'s difficulty in building non-OCaml software. I am not very motivated to spend time supporting this approach, and I was sort of hoping that dune would eventually choose a different path.

The PR itself is supporting relocatability, which (solves dune's problem and) is something that I find reasonable in general. The thing that would give me pause is that we keep adding layers and layers of complexity to this logic, and we never remove or simplify the previous layers away. (We are not motivated to simplify the code because we want maintenance-only changes and simplifications have a risk of regressions, but the result of this approach is to increase complexity and maintenance harder over time.)

Is there a "clean" approach to write relocatable OCaml programs these day, even if they perform some sort of dynamic linking? (ocamlbuild needs to locate its own libraries at runtime to build plugins.) The PR implements the approach of guessing the libdir from the bindir, going to `../lib/ocamlbuild` from the location of the binary; this works for the standard `opam` install layout, but I would worry about other filesystem organizations (for example `ocamlfind` by default puts things in a `/lib/ocaml/site-lib/` subdirectory). Does the dune sandbox environment respect this convention?

Calling `ocamlfind query ocamlbuild` at runtime may be the easiest way to find this out, because giving library directories for packages is precisely its job -- but again, is the dune sandbox environment correctly configured to make this work before ocamlbuild is actually installed?

hhugo:
Related to https://github.com/dra27/ocamlbuild/pull/2

gasche:
That is a different PR from @dra27 that also makes ocamlbuild relocatable, on top of his work to make the OCaml compiler itself relocatable. I asked David a few days ago (during the github/ocaml triaging meeting) about this general question of how to make OCaml programs relocatable, and my worry that assuming a fixed relative path from bindir to libdir is not expressive enough to work with diverse packaging choices. For OCaml, David made the bindir->libdir relative path configurable (at configure-time), which is flexible enough to capture many use-cases. This would probably also be a fine approach for ocamlbuild -- provide a configure-time choice for the bindir->libdir relative path.

My impression is that this idea of configurable relative bindir->libdir path should be enough to support the systems that we currently support well, possibly with an explicit opt-out to set libdir at configure and then at runtime to cover weird cases (I'm not sure how things work on Nix). The difference relative to the present PR would not be very large -- but I would be interested in trying to remove some of the older stuff to simplify our lives.

(The patch of David does something different, using new primitives of the compiler introduced by the Relocatable PRs, and I think it deserves its own design explanation and discussion.)

gasche:
The Relocatable OCaml PR has been merged, so it may be time to revisit this discussion. @dra27, @gridbugs, which approach to relocatability should we consider at the level of ocamlbuild? Since the initial discussion I have mellowed out on the idea of doing a relative lookup; but maybe David's approach of making this relative-path configurable at configure-time would be better. And then David's PR at https://github.com/dra27/ocamlbuild/pull/2 is doing things fairly differently.

I would like to help getting ocamlbuild to be relocatable now that the compiler also is, but what is the right approach to do this?

Alizter:
I wasn't following this issue before, but we are trying to get ocamlbuild working with the new relocatable branches for the compiler and ocamlbuild. We are tracking the progress here https://github.com/ocaml/dune/issues/13229.

AFAIK @gridbugs is no longer working on this.

gasche:
I think it would be useful to have @dra27's perspective on this PR. Given that I have otherwise mellowed out on the proposed implementation choice, I may support just merging it to enable relocatability -- maybe without the warning, if we decide that relocation is okay.

Alizter:
Just to be clear, I don't support merging this PR, at least not for the reason of compatibility with dune pkg.

Speaking as a dune maintainer, I am more in favour of supporting the existing packaging until there is something really wrong with how something is packaged. I therefore don't think this PR is really a good fit for purpose at the moment.

@dra27 has his own relocatable ocamlbuild patch https://github.com/dra27/ocamlbuild/commit/1b4b2ea49762cda69b288616429e83535f7048cf.patch?full_index=1 that we are currently testing out for dune. Ideally that should be merged here when deemed fit-for-purpose in the guise of increasing the relocatilbity of ocamlbuild. In our own testing of ocamlbuild packages + dune pkg that we are working on, we will likely uncover any further issues that will need to be communicated in this repo.

gasche:
> I don't support merging this PR, at least not for the reason of compatibility with dune pkg.

I was previously grumpy about the reported use-case, but then I do agree that relocatability is nice -- and especially more relevant now that the compiler itself has become relocatable, so that it's tempting to push for the whole ecosystem to be relocatable. With this new expanded use-case, the PR proposed is something I find more tempting than before. (But I agree that asking for @dra27's opinion is a reasonable intermediate step, if he has time to comment.)

PR #328, opened 2024-05-16
Setup github actions for ocamlbuild

hhugo:
This PR migrate the existing travis.yml CI instructions to github-actions.
Currently testing OCaml 4.08 .. 5.2

It also includes the following fixes.

- fix testsuite (see https://github.com/ocaml/ocamlbuild/issues/321)
- fix bootstrap
- fix tests for Ocaml 5.2

The underlying reason for the MR is to make is easier to test fixes for windows, such as https://github.com/ocaml/ocamlbuild/pull/323

### DIFF
--- a/testsuite/ocamlbuild_test.ml
+++ b/testsuite/ocamlbuild_test.ml
@@ -435,19 +435,51 @@ let print_colored header_color header name body_color body =
     (color_code header_color) header name
     (color_code body_color) body
 
+
 let run ~root =
   let dir = Sys.getcwd () in
   let root = dir ^ "/" ^ root in
   rm root;
   Unix.mkdir root 0o750;
+  let install_lib_dir = Filename.concat root "install-lib" in
+  let install_bin_dir = Filename.concat root "install-bin" in
 
   let test_tree = Filename.dirname (Sys.argv.(0)) in
   let test_tree = if test_tree = "." then Sys.getcwd () else test_tree in
   let build_tree = Filename.dirname test_tree in
-  let ocamlbuild = Filename.concat build_tree "ocamlbuild.byte" in
+
+  let copy l dest =
+    ignore(Sys.command (Printf.sprintf "mkdir -p %s" dest));
+    List.iter (fun f -> ignore(Sys.command (Printf.sprintf "cp %s/%s %s" build_tree f dest))) l
+  in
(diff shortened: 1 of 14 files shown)


hhugo:
also, replaces #317

testsuite/ocamlbuild_test.ml
```diff
@@ -435,19 +435,51 @@ let print_colored header_color header name body_color body =
...
+  let install_bin_dir = Filename.concat root "install-bin" in
 
   let test_tree = Filename.dirname (Sys.argv.(0)) in
   let test_tree = if test_tree = "." then Sys.getcwd () else test_tree in
   let build_tree = Filename.dirname test_tree in
-  let ocamlbuild = Filename.concat build_tree "ocamlbuild.byte" in
+
+  let copy l dest =
+    ignore(Sys.command (Printf.sprintf "mkdir -p %s" dest));
+    List.iter (fun f -> ignore(Sys.command (Printf.sprintf "cp %s/%s %s" build_tree f dest))) l
```
gasche, line 453: Could you write these with `Sys.command @@ Filename.quote_command ...`? It is good practice to protect ourselves against spaces in the build path.

gasche (approved):
This is impressive work, thanks!

testsuite/ocamlbuild_test.ml
hhugo: `Filename.quote_command` was introduced in OCaml 4.10. I could use `Filename.quote` but I don't really understand the implication of its documentation
```
    Warning: under Windows, the output is only suitable for use
    with programs that follow the standard Windows quoting
    conventions.
```

gasche: (Debian stable has 4.13, oldstable has 4.11.)

hhugo: We probably need to bump the constraint in the opam file. It's currently 4.03, I think

gasche:
@hhugo would you be interested in the commit bit on ocamlbuild? I am not actively maintaining it, I guess that it could benefit from some help, and you seem very dedicated.

hhugo:
> @hhugo would you be interested in the commit bit on ocamlbuild? I am not actively maintaining it, I guess that it could benefit from some help, and you seem very dedicated.

I can do a bit of support, you can add me.

PR #329, opened 2024-05-27
Enable CI on windows and fix it.

### LINKED ISSUE #316: Windows/Cygwin: hygienic check fails for some packages
It looks like an [old issue](https://github.com/ocaml/opam-repository/issues/6224) has popped up again in `ocamlbuild` `0.14.1`.
For some packages on Windows/Cygwin (Mingw toolchain), the hygienic check fails, e.g., for `mtime` `1.3.0`:
```
mkdir 'C:\Opt\OCaml64\home\kayuwe.kirstein\.opam\4.13.1+mingw64\.opam-switch\build\mtime.1.3.0\_build'
Das System kann den angegebenen Pfad nicht finden.
''ocamlfind ocamlopt unix.cmxa -I 'C:/Opt/OCaml64/home/kayuwe.kirstein/.opam/4.13.1+mingw64/lib\ocamlbuild' 'C:/Opt/OCaml64/home/kayuwe.kirstein/.opam/4.13.1+mingw64/lib\ocamlbuild/ocamlbuildlib.cmxa' -linkpkg myocamlbuild.ml 'C:/Opt/OCaml64/home/kayuwe.kirstein/.opam/4.13.1+mingw64/lib\ocamlbuild/ocamlbuild.cmx' -o myocamlbuild.exe
SANITIZE: a total of 3 files that should probably not be in your source tree
  has been found. A script shell file
  "C:\\Opt\\OCaml64\\home\\kayuwe.kirstein\\.opam\\4.13.1+mingw64\\.opam-switch\\build\\mtime.1.3.0\\_build/sanitize.sh"
  is being created. Check this script and run it to remove unwanted files or
  use other options (such as defining hygiene exceptions or using the
  -no-hygiene option).
IMPORTANT: I cannot work with leftover compiled files.
ERROR: Leftover object files:
  File myocamlbuild.o in _build has suffix .o
ERROR: Leftover OCaml compilation files:
  File myocamlbuild.cmi in _build has suffix .cmi
  File myocamlbuild.cmx in _build has suffix .cmx
Exiting due to hygiene violations.
pkg.ml: [ERROR] cmd ["ocamlbuild" "-use-ocamlfind" "-classic-display" "-j" "4" "-tag" "debug"
     "-build-dir" "_build" "opam" "pkg/META" "CHANGES.md" "LICENSE.md"
     "README.md" "src/mtime.a" "src/mtime.cmxs" "src/mtime.cmxa"
     "src/mtime.cma" "src/mtime.cmx" "src/mtime.cmi" "src/mtime.mli"
     "src/mtime_top.a" "src/mtime_top.cmxs" "src/mtime_top.cmxa"
     "src/mtime_top.cma" "src/mtime_top.cmx" "src/mtime_top_init.ml"
     "doc/index.mld" "src-os/mtime_clock.a" "src-os/mtime_clock.cmxs"
     "src-os/mtime_clock.cmxa" "src-os/mtime_clock.cma"
     "src-os/mtime_clock.cmx" "src-os/mtime_clock.cmi"
     "src-os/mtime_clock.mli" "src-os/dllmtime_clock_stubs.dll"
     "src-os/libmtime_clock_stubs.a" "test-os/min_os.ml"]: exited with 1
```
I am not sure whether the first error message on not finding the build folder is a serious failure.
Executing `sanitize.sh` does not help, I guess `opam install mtime` regenerates the object files every time.

Observed on Windows 10, Cygwin/Mingw (Ocaml for Windows) for OCaml `4.13.1+mingw64`.
My current workaround is downgrading `ocamlbuild` to `0.14.0`.

hhugo:
This is a somewhat minimal list of changes to make the testsuite pass on windows.
If you look at the CI jobs, you should see that the testsuite pass under windows.

Includes:
- part of #70 https://github.com/ocaml/ocamlbuild/pull/70/commits/62c40ca26173f794790243d608d6415d0a64a6ce
- part of #323 

Probably fix #316

### DIFF
--- a/testsuite/internal.ml
+++ b/testsuite/internal.ml
@@ -184,21 +184,21 @@ let () = test "CmxsStubLink"
            T.f "libfoo_stubs.clib" ~content:"foo_stubs.o";
            T.f "foo.ml" ~content:"";
          ];
-         T.f "_tags" ~content:"
-<src/foo.{cma,cmxa}> : record_foo_stubs
-<src/foo.cmxs> : link_foo_stubs";
-         T.f "myocamlbuild.ml" ~content:"
-open Ocamlbuild_plugin
-let () =
-  dispatch begin function
-  | After_rules ->
-      dep [\"record_foo_stubs\"] [\"src/libfoo_stubs.a\"];
-      flag_and_dep
-        [\"link\"; \"ocaml\"; \"link_foo_stubs\"] (P \"src/libfoo_stubs.a\");
-      flag [\"library\"; \"ocaml\";           \"record_foo_stubs\"]
-        (S ([A \"-cclib\"; A \"-lfoo_stubs\"]));
-  | _ -> ()
-  end
+         T.f "_tags" ~content:"\
+<src/foo.{cma,cmxa}> : record_foo_stubs\n\
+<src/foo.cmxs> : link_foo_stubs\n";
+         T.f "myocamlbuild.ml" ~content:"\
+open Ocamlbuild_plugin\n\
+let () =\n\
+  dispatch begin function\n\
+  | After_rules ->\n\
+      dep [\"record_foo_stubs\"] [\"src/libfoo_stubs.a\"];\n\
+      flag_and_dep\n\
+        [\"link\"; \"ocaml\"; \"link_foo_stubs\"] (P \"src/libfoo_stubs.a\");\n\
+      flag [\"library\"; \"ocaml\";           \"record_foo_stubs\"]\n\
+        (S ([A \"-cclib\"; A \"-lfoo_stubs\"]));\n\
+  | _ -> ()\n\
+  end\n\
 "]
   ~targets:("src/foo.cmxs",[]) ();;
 
@@ -208,31 +208,31 @@ let () = test "StrictSequenceFlag"
   ~tree:[T.f "hello.ml" ~content:"let () = 1; ()";
          T.f "_tags" ~content:"true: strict_sequence\n"]
   ~failing_msg:(if Sys.ocaml_version < "4.07.0" then
-"File \"hello.ml\", line 1, characters 9-10:
-Error: This expression has type int but an expression was expected of type
-         unit
+"File \"hello.ml\", line 1, characters 9-10:\n\
+Error: This expression has type int but an expression was expected of type\n\
+\         unit\n\
 Command exited with code 2."
 else if Sys.ocaml_version < "4.08.0" then
-"File \"hello.ml\", line 1, characters 9-10:
-Error: This expression has type int but an expression was expected of type
-         unit
-       because it is in the left-hand side of a sequence
+"File \"hello.ml\", line 1, characters 9-10:\n\
+Error: This expression has type int but an expression was expected of type\n\
+\         unit\n\
+\       because it is in the left-hand side of a sequence\n\
 Command exited with code 2."
 else if Sys.ocaml_version < "5.2.0" then
-"File \"hello.ml\", line 1, characters 9-10:
-1 | let () = 1; ()
-             ^
-Error: This expression has type int but an expression was expected of type
-         unit
-       because it is in the left-hand side of a sequence
+"File \"hello.ml\", line 1, characters 9-10:\n\
+1 | let () = 1; ()\n\
+\             ^\n\
+Error: This expression has type int but an expression was expected of type\n\
+\         unit\n\
+\       because it is in the left-hand side of a sequence\n\
 Command exited with code 2."
 else
-"File \"hello.ml\", line 1, characters 9-10:
-1 | let () = 1; ()
-             ^
-Error: This expression has type \"int\" but an expression was expected of type
-         \"unit\"
-       because it is in the left-hand side of a sequence
+"File \"hello.ml\", line 1, characters 9-10:\n\
+1 | let () = 1; ()\n\
+\             ^\n\
+Error: This expression has type \"int\" but an expression was expected of type\n\
+\         \"unit\"\n\
+\       because it is in the left-hand side of a sequence\n\
 Command exited with code 2."
 )
   ~targets:("hello.byte",[]) ();;
@@ -243,16 +243,16 @@ let () = test "StrictFormatsFlag"
   ~tree:[T.f "hello.ml" ~content:"let _ = Printf.printf \"%.10s\"";
          T.f "_tags" ~content:"true: strict_formats\n"]
   ~failing_msg:(if Sys.ocaml_version < "4.08.0" then
-"File \"hello.ml\", line 1, characters 22-29:
+"File \"hello.ml\", line 1, characters 22-29:\n\
 Error: invalid format \"%.10s\": at character number 0, \
-`precision' is incompatible with 's' in sub-format \"%.10s\"
+`precision' is incompatible with 's' in sub-format \"%.10s\"\n\
 Command exited with code 2."
 else
-"File \"hello.ml\", line 1, characters 22-29:
-1 | let _ = Printf.printf \"%.10s\"
-                          ^^^^^^^
+"File \"hello.ml\", line 1, characters 22-29:\n\
+1 | let _ = Printf.printf \"%.10s\"\n\
+\                          ^^^^^^^\n\
 Error: invalid format \"%.10s\": at character number 0, \
-`precision' is incompatible with 's' in sub-format \"%.10s\"
+`precision' is incompatible with 's' in sub-format \"%.10s\"\n\
 Command exited with code 2."
 )
   ~targets:("hello.byte",[]) ();;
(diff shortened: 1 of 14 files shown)


testsuite/internal.ml
```diff
@@ -182,21 +184,21 @@ let () = test "CmxsStubLink"
...
+let () =\n\
+  dispatch begin function\n\
+  | After_rules ->\n\
+      dep [\"record_foo_stubs\"] [\"src/libfoo_stubs.a\"];\n\
+      flag_and_dep\n\
+        [\"link\"; \"ocaml\"; \"link_foo_stubs\"] (P \"src/libfoo_stubs.a\");\n\
+      flag [\"library\"; \"ocaml\";           \"record_foo_stubs\"]\n\
+        (S ([A \"-cclib\"; A \"-lfoo_stubs\"]));\n\
+  | _ -> ()\n\
+  end\n\
```
gasche, line 201: Urgh, can you use quoted literals instead to avoid the many escapes here? (If there is a problem with `\r\n` somehow, we could inject a small helper function that removes the `\r`.)

src/my_std.ml
```diff
@@ -279,7 +279,7 @@ let sys_command =
   match Sys.os_type with
   | "Win32" -> fun cmd ->
       if cmd = "" then 0 else
-      let cmd = "bash --norc -c " ^ Filename.quote cmd in
+      let cmd = "bash --norc -c " ^ Filename.quote cmd ^ " 2>&1" in
```
gasche, line 282: Can you explain why we need this change?

gasche (commented):
This looks okay overall, see one question on the stderr rediction.

I'm not fond of the generalized use of explicit newline quoting in string literals in the testsuite, I find that it really hurts readability. Can we use quoted string literals instead?

testsuite/internal.ml
hhugo: I've pushed https://github.com/ocaml/ocamlbuild/pull/329/commits/0f2342a2236c60252b638617c9c11752a7f6189c

src/my_std.ml
hhugo: I'm not happy with this change but it makes the testsuite pass.

I think the explanation is as follow:
- if `is_degraded` or Sys`.win32` is true, `execute_many` falls back to `sys_command` to implement `execute_many`.
- otherwise, the logic from `Ocamlbuild_executor.execute` is used, which reads both stderr and stdout into a `job_buffer`.
- The testuite sometime expect some output on stdout.  

I'm still looking into a better solution.

hhugo: I've moved the stderr redirection to `command.ml` where I think it belongs and added a small comment.

gasche: I went to look at the code of ocamlbuild_executor to understand what is going on. `ocamlbuild_executor` does not just redirect stderr into stdout, it lets the command write to different places *and then* on job termination it adds stdout, then stderr, to the job buffer, in that order. The change that you implemented does not match this behavior, it only works for commands that have only stdout or only stderr output.

gasche: Out of curiosity: in the `execute_many` definition, have you tried removing the "or we are on Windows" condition, to get the ocamlbuild_executor logic to run also on Windows? When ocamlbuild was first written, I think the Windows support in stdlib and Unix was weaker than it is now, there is a small chance that removing those alternate paths could work today.

(We could probably also get rid of the "is_degraded" notion, which I think was designed to let ocamlbuild build the compiler in constrained environments. ocamlbuild is not considered as build system for the OCaml compiler anymore, so maybe we could forget about this case altogether. But this is more work than I was considering putting in ocamlbuild again.)

src/shell.ml
```diff
@@ -39,7 +39,7 @@ let run args target =
   let cmd = String.concat " " (List.map quote_filename_if_needed args) in
   if !*My_unix.is_degraded || Sys.os_type = "Win32" then
     begin
-      Log.event cmd target Tags.empty;
+      if false then Log.event cmd target Tags.empty;
```
kit-ty-kate, line 42: Why is this disabled?
```suggestion
      Log.event cmd target Tags.empty;
```

testsuite/ocamlbuild_test.ml
```diff
@@ -435,6 +439,17 @@ let print_colored header_color header name body_color body =
...
+  if not (Sys.file_exists dir)
+  then
+    try Unix.mkdir dir 0o750 with _ -> ()
+
+let rec mkdir_p dir =
+  if Sys.file_exists dir then ()
+  else begin
+      mkdir_p (Filename.dirname dir);
+      try_mkdir dir
+    end
```
kit-ty-kate, line 452:
```suggestion
let rec mkdir_p dir =
  if not (Sys.file_exists dir) then begin
    mkdir_p (Filename.dirname dir);
    Unix.mkdir dir 0o750;
  end
```

testsuite/ocamlbuild_test.ml
```diff
@@ -32,10 +32,14 @@ let print_string_list_blank = print_list_blank pp_print_string
 
 let exists filename =
   try ignore(Unix.stat filename); true
-  with Unix.Unix_error ((Unix.ENOENT),_,_) -> false
+  with Unix.Unix_error ((Unix.ENOENT),_,_) ->
+    if Sys.win32
+    then Sys.file_exists filename
+    else false
```
kit-ty-kate, line 38: why not simply use `Sys.file_exists` on all platforms?
```suggestion
  Sys.file_exists filename
```

kit-ty-kate (commented):
Looks fine on first glance with my minimal Windows experience.

cc-ing more experienced Windows users @dra27 @jonahbeckford

src/shell.ml
hhugo: It make the testsuite fail because it output `cmd` on stdout.
I wonder if we should prefix the command with `"+ "` similar to what is done in non-degraded mode. The testsuite filters theses lines out already.

testsuite/ocamlbuild_test.ml
hhugo: I don't know how the behavior defer, I wanted to avoid changing the behavior in non windows as much as possible. But it's probably fine inside the testsuite.

hhugo: applied

src/shell.ml
hhugo: I've added the `+ ` prefix

src/my_std.ml
hhugo: > I went to look at the code of ocamlbuild_executor to understand what is going on. `ocamlbuild_executor` does not just redirect stderr into stdout, it lets the command write to different places _and then_ on job termination it adds stdout, then stderr, to the job buffer, in that order. The change that you implemented does not match this behavior, it only works for commands that have only stdout or only stderr output.

Looking at the code quickly, my impression is that it can interleave stdout and stderr. In the main `loop`, you can see
```
chrfds, do_read ~loop:false;
```
which tries to read any both stdout and stderr and push the content to `job_buffer`.

Maybe I'm misunderstanding the logic.

gasche: Ok, this suggests that I misunderstood, and there may be no good way to ensure that we get the same output. In that context I am okay with your current choice even though it is sort of a hack.

gasche:
Ideally what I would like to do is to get good Windows support without making the codebase worse than it already is. I think that the current PR you propose is okay in this respect, the changes are relatively minimal and at the same time justified, and the risk of regression for non-Windows users is low. I propose to merge it and build up from there.

On the other hand I am more worried about the fdopen patches in #330, which look like a mix of good ideas (I like getting rid of the "or is win32" case in the "is degraded?" test), inscrutable hacks, and a things in the middle, and was never really tested on non-Windows systems if I understand correctly. Taking the good pieces out of it is an excellent idea, and maybe all of it is good and I just don't understand it yet. But making our codebase harder to understand (obscure changes with little explanations and no commit message) is getting convenience right now but also even more pain later, so I would not be in favor.

Note: I noticed the same `Be consistent!` comment in the present PR and in the .patch file somewhere. Are you the author of all the changes in the current PR? If not, are reused fragment properly credited and do they have a compatible software license?

gasche (approved):

hhugo:
https://github.com/ocaml/ocamlbuild/pull/329/commits/00b92ab0f67698e3d85be72f02c68d00a3374ec6 was extracted from an old PR  #70 -> https://github.com/ocaml/ocamlbuild/pull/70/commits/62c40ca26173f794790243d608d6415d0a64a6ce

gasche:
Could you cherry-pick that commit (or amend the author information on your own commit) to properly credit fdopen in the metadata?

kit-ty-kate:
> @kit-ty-kate are you okay with merging this? My reasoning is: the patch looks okay, the potential for non-Windows regression is minimal, and I want to make it as easy as reasonably possible for @hhugo (or of course anyone else) to improve the state of ocamlbuild on Windows.

Merging is fine as long as we remember to not release before getting the double-check/approval of the two people i've cc'd above. I'd like to especially double-check the change made to `plugin-lib/ocamlbuild_unix_plugin.ml` and `src/command.ml`, which seem a bit fragile, especially as:
- `bash` could be present in the environment, without it being linked to an MSYS2 or Cygwin installation (e.g. some programs like some git installation ship with `bash.exe`)
- I'm not certain the quoting used is always correct with cmd.exe but I may be wrong. In particular I would like to understand this comment from 8 years ago: https://github.com/ocaml/ocamlbuild/pull/70#issuecomment-223121978
- Calls to `bash` uses `--noprofile` in https://github.com/ocaml/ocamlbuild/pull/330, is there a reason why this has not been done here too?

src/my_std.ml
hhugo: > Out of curiosity: in the `execute_many` definition, have you tried removing the "or we are on Windows" condition, to get the ocamlbuild_executor logic to run also on Windows? When ocamlbuild was first written, I think the Windows support in stdlib and Unix was weaker than it is now, there is a small chance that removing those alternate paths could work today.
> 
> (We could probably also get rid of the "is_degraded" notion, which I think was designed to let ocamlbuild build the compiler in constrained environments. ocamlbuild is not considered as build system for the OCaml compiler anymore, so maybe we could forget about this case altogether. But this is more work than I was considering putting in ocamlbuild again.)

Done in #333 and #332

PR #333, opened 2024-05-29
Use normal codepath for win32

hhugo:
The PR removes the "degraded" code path when `Sys.win32` is `true`.

It goes in the direction of #330 but leaves a bunch of diffs out for now.

- It doesn't normalize path '/' vs `\`
- It doesn't search for a shell (bash.exe, dash.exe, sh.exe) and unconditionally uses bash.
- It doesn't protect against command line length limit 
- It doesn't contain the following "fix" (not sure what the bug is ...)
```
let prep_windows_cmd cmd =
  (* workaround known ocaml bug, remove later *)
  if String.contains cmd '\t' && String.contains cmd ' ' = false then
    " " ^ cmd
  else
    cmd
```

### DIFF
--- a/src/ocamlbuild_executor.ml
+++ b/src/ocamlbuild_executor.ml
@@ -73,7 +73,10 @@ let output_lines prefix oc buffer =
         try String.index_from u i '\n'
         with Not_found -> m
       in
-      output_line i j;
+      (* ignore trailing CR *)
+      let k = ref j in
+      while !k > i && u.[!k - 1] = '\r' do decr k done;
+      output_line i !k;
       loop (j + 1)
     else
       ()
--- a/src/ocamlbuild_executor.ml
+++ b/src/ocamlbuild_executor.ml
@@ -133,11 +133,18 @@ let execute
   (* ***)
   (*** add_job *)
   let add_job cmd rest result id =
+    let cmd =
+      if Sys.win32
+      then "bash --norc -c " ^ Filename.quote cmd
+      else cmd
+    in
     (*display begin fun oc -> fp oc "Job %a is %s\n%!" print_job_id id cmd; end;*)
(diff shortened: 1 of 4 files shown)


hhugo:
One should probably wait for #336 before merging so that we can see that it works without any hacks from the sunset repo.

hhugo:
I've merged #330, this is now ready for review. cc @kit-ty-kate @gasche

src/ocamlbuild_executor.ml
```diff
@@ -73,7 +73,10 @@ let output_lines prefix oc buffer =
         try String.index_from u i '\n'
         with Not_found -> m
       in
-      output_line i j;
+      (* ignore trailing CR *)
+      let k = ref j in
+      while !k > i && u.[!k - 1] = '\r' do decr k done;
```
kit-ty-kate, line 78: to avoid polymorphic comparison
```suggestion
      while !k > (i : int) && u.[!k - 1] = '\r' do decr k done;
```

src/shell.ml
```diff
@@ -24,33 +24,36 @@ let is_simple_filename s =
...
+
+(*** Copied from ocaml/stdlib/filename.ml *)
+let generic_quote quotequote s =
+  let l = String.length s in
+  let b = Buffer.create (l + 20) in
+  Buffer.add_char b '\'';
+  for i = 0 to l - 1 do
+    if s.[i] = '\''
+    then Buffer.add_string b quotequote
+    else Buffer.add_char b  s.[i]
```
kit-ty-kate, line 36:
```suggestion
    else Buffer.add_char b s.[i]
```

kit-ty-kate (approved):
apart from the couple of comments below that looks fine to me

PR #335, opened 2024-05-30
Only support OCaml versions tested in the CI

hhugo:


gasche:
Debian stable is on 4.13.1, debian oldstable is on 4.11.1. Updating from 4.03 to 4.08 sounds very reasonable to me (and any thing up to 4.11 would be fine as well). I think that we *could* keep support for 4.03 without too much cost, but there are also small benefits to upgrading the lower bound due to stdlib improvements.

PR #339, opened 2024-06-20
No longer call into cmd.exe to execute a posix shell on windows

hhugo:
based on top of #338 
Inspired by #330.
`Sys.command`, `Unix.open_process_in`, `Unix.open_process_full` all go through `cmd.exe` to execute the given command.
In our case, it means we call cmd.exe to call bash.
This is unnecessary, and complicate things. We can directly call a posix shell and not have to worry about another layer of escaping.

In this PR, we use `open_process_args*` funtions to avoid going through `cmd.exe`

### DIFF
--- a/src/my_std.ml
+++ b/src/my_std.ml
@@ -291,6 +291,9 @@ let split_path_win str =
   in
   aux 0
 
+(* Here to break the circular dep *)
+let log3 = ref (fun _ -> failwith "My_std.log3 not initialized")
+
 let windows_shell = lazy begin
   let rec iter = function
   | [] -> [| "bash.exe" ; "--norc" ; "--noprofile" |]
@@ -305,7 +308,7 @@ let windows_shell = lazy begin
   in
   let paths = split_path_win (try Sys.getenv "PATH" with Not_found -> "") in
   let res = iter paths in
-  if false then Printf.eprintf "Using shell %s\n%!" (Array.to_list res |> String.concat " ");
+  !log3 (Printf.sprintf "Using shell %s" (Array.to_list res |> String.concat " "));
   res
 end
 
--- a/src/my_std.ml
+++ b/src/my_std.ml
@@ -301,7 +301,7 @@ let log3 = ref (fun _ -> failwith "My_std.log3 not initialized")
 
 let windows_shell = lazy begin
   let rec iter = function
-  | [] -> [| "bash.exe" ; "--norc" ; "--noprofile" |]
+  | [] -> raise Not_found
   | hd::tl ->
     let dash = Filename.concat hd "dash.exe" in
     if Sys.file_exists dash then [|dash|] else
@@ -312,9 +312,18 @@ let windows_shell = lazy begin
     if Sys.file_exists sh then [|sh|] else [|bash ; "--norc" ; "--noprofile"|]
   in
   let paths = split_quoted (try Sys.getenv "PATH" with Not_found -> "") ';' in
-  let res = iter paths in
-  !log3 (Printf.sprintf "Using shell %s" (Array.to_list res |> String.concat " "));
-  res
+  let shell =
+    try
+      let path =
+        List.find (fun path ->
+            Sys.file_exists (Filename.concat path "cygcheck.exe")) paths
+      in
+      iter [path]
+    with Not_found ->
+      (try iter paths with Not_found -> failwith "no posix shell found in PATH")
+  in
+  !log3 (Printf.sprintf "Using shell %s" (Array.to_list shell |> String.concat " "));
+  shell
 end
 
 let prepare_command_for_windows cmd =
--- a/src/my_std.ml
+++ b/src/my_std.ml
@@ -327,6 +327,9 @@ let windows_shell = lazy begin
 end
 
 let prepare_command_for_windows cmd =
+  (* The best way to prevent bash from switching to its windows-style
+   * quote-handling is to prepend an empty string before the command name. *)
+  let cmd = "''" ^ cmd in
   Array.append (Lazy.force windows_shell) [|"-c"; cmd|]
 
 let sys_command_win32 cmd =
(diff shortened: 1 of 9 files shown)


src/my_std.ml
```diff
@@ -275,13 +275,54 @@ let sys_file_exists x =
...
+  if cmd = "" then 0 else
+  if Sys.win32
+  then
+    let args = prepare_command_for_windows cmd in
+    let oc = Unix.open_process_args_out args.(0) args in
+    match Unix.close_process_out oc with
+    | WEXITED x -> x
+    | WSIGNALED _ -> 2 (* like OCaml's uncaught exceptions *)
+    | WSTOPPED _ -> 127
+  else Sys.command cmd
```
dra27, line 325: Worth splitting at the toplevel?
```suggestion
let sys_command_win32 cmd =
  let args = prepare_command_for_windows cmd in
  let oc = Unix.open_process_args_out args.(0) args in
  match Unix.close_process_out oc with
  | WEXITED x -> x
  | WSIGNALED _ -> 2 (* like OCaml's uncaught exceptions *)
  | WSTOPPED _ -> 127

let sys_command =
  if Sys.win32 then
    sys_command_win32
  else
    Sys.command

let sys_command cmd =
  if cmd = "" then 0 else
  sys_command cmd
```

dra27 (commented):
Couple of suggestions, but this looks very sensible. I'd definitely recommend the `cygcheck` trick - if `ocamlbuild` is run from cmd/powershell then by default bash is often going to be WSL, so it may be a good fallback ocamlbuild just to fail at that point (rather than run WSL's bash and the results be very surprising).

When ocamlbuild runs within an opam package (which is where it should be running 99% of the time), everything'll work as expected.

src/my_std.ml
```diff
@@ -275,13 +275,54 @@ let sys_file_exists x =
...
+      let len = String.length str - pos in
+      if len = 0 then [] else [String.sub str pos len]
+  in
+  aux 0
+
+let windows_shell = lazy begin
+  let rec iter = function
+  | [] -> [| "bash.exe" ; "--norc" ; "--noprofile" |]
+  | hd::tl ->
+    let dash = Filename.concat hd "dash.exe" in
```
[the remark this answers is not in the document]
hhugo: Are you suggesting 
1. to only do the cygcheck trick and fail if not found or
2. to fallback to the current logic ?

src/my_std.ml
```diff
@@ -275,13 +275,54 @@ let sys_file_exists x =
...
+    if Sys.file_exists dash then [|dash|] else
+    let bash = Filename.concat hd "bash.exe" in
+    if not (Sys.file_exists bash) then iter tl else
+    (* if sh.exe and bash.exe exist in the same dir, choose sh.exe *)
+    let sh = Filename.concat hd "sh.exe" in
+    if Sys.file_exists sh then [|sh|] else [|bash ; "--norc" ; "--noprofile"|]
+  in
+  let paths = split_path_win (try Sys.getenv "PATH" with Not_found -> "") in
+  let res = iter paths in
+  if false then Printf.eprintf "Using shell %s\n%!" (Array.to_list res |> String.concat " ");
```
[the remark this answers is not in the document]
hhugo: There is but there a circular dep issue. I've added the logic to work around it now

src/my_std.ml
```diff
@@ -275,13 +275,54 @@ let sys_file_exists x =
...
+      let len = i - pos in
+      if len = 0 then
+        aux (succ i)
+      else
+        String.sub str pos (i - pos) :: aux (succ i)
+    with Not_found | Invalid_argument _ ->
+      let len = String.length str - pos in
+      if len = 0 then [] else [String.sub str pos len]
+  in
+  aux 0
```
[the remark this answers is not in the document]
hhugo: Done, thanks

src/my_std.ml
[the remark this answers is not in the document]
hhugo: I implemented 2. for now

src/my_std.ml
```diff
@@ -275,13 +275,80 @@ let sys_file_exists x =
       try Array.iter (fun x -> if x = basename then raise Exit) a; false
       with Exit -> true
 
+(* https://github.com/ocaml/opam/blame/master/src/core/opamStd.ml *)
```
gasche, line 278: (Luckily opam uses the same license so reusing code as-is is just fine.)

@hhugo could you maybe tweak your URL to contain a hash of a recent commit, instead of `master`? This would be helpful if someone in the future wants to now whether our local copy should be updated.

(I would have considered moving the copied-almost-exactly-as-is code to a separate file, to make such future tracking easier, but oh well.)

gasche (commented):
I haven't read the Windows details in details, but I agree with the idea of reusing tricky code from opam, so I'm in favor. (`prepare_command_for_windows` is a nice abstraction.) See minor review comment.

hhugo:
There was some logic elsewhere to read the PATH (that was not handling quoted path).
I've unified things in https://github.com/ocaml/ocamlbuild/pull/339/commits/d834dd075fd257037d9e919a7d1172034c2b9990

src/my_std.ml
hhugo: I've updated the comment

src/ocamlbuild_executor.ml
```diff
@@ -136,13 +136,13 @@ let execute
...
-      then "bash --norc -c " ^ Filename.quote cmd
-      else cmd
-    in
     (*display begin fun oc -> fp oc "Job %a is %s\n%!" print_job_id id cmd; end;*)
-    let (stdout', stdin', stderr') = open_process_full cmd env in
+    let (stdout', stdin', stderr') =
+      if Sys.win32
+      then
+        let args = My_std.prepare_command_for_windows cmd in
+        open_process_args_full args.(0) args (Unix.environment ())
```
gasche, line 144: Why are you calling `Unix.environment ()` again here, what is the expected difference with just `env`?

hhugo: no reason, I've changed it back.

gasche (approved):

Issue #322, opened 2023-09-19
Can we have a unified opam file next version?


jonahbeckford:
It is cumbersome to have one `ocamlbuild.0.14.2/opam` version for Unix and another `ocamlbuild.0.14.2+win/opam` for Windows with availability flags to choose between both.

Among other things, we've lost the ability to pin ocamlbuild to a specific version.

I unified them in my own custom repository at https://github.com/diskuv/diskuv-opam-repository/tree/main/packages/ocamlbuild/ocamlbuild.0.14.2%2Bwin%2Bunix . Can we consider doing the same when it comes time to publish ocamlbuild.0.14.3? Thanks.

kit-ty-kate:
For the record afaik none of the ocamlbuild maintainers were for the split. If I weren't undergoing a severe burnout at the time I would've never accepted this in the repository.

As far as I know we have not received any proper PR on this here repository for this yet and your contribution would be most welcome, thanks a lot!

see https://github.com/ocaml/opam-repository/pull/23832 for the original addition

hhugo:
https://github.com/ocaml/opam-repository/pull/26164

PR #340, opened 2024-06-20
Update changelog

hhugo:
on top of #334 #344

### DIFF
--- a/Changes
+++ b/Changes
@@ -13,7 +13,21 @@ of the usual "- ".
 
 NEXT_RELEASE:
 -------------
-
+- Misc: restore CI (unix and windows)
+  (#328, #329, #336 by Hugo Heuzard)
+* Remove degraded mode for windows
+  (#333 by Hugo Heuzard)
+- Remove light mode (#332 by Hugo Heuzard)
+- Make the codebase work on windows native
+  (#329, #333, #334, #338, #339, #342, #343, #344 by Hugo Heuzard)
+* No longer treat empty path in PATH env variable as the current working directory
+  (#339 by Hugo Heuzard)
+- Emit a warning if several calls are made to
+   `Ocamlbuild_plugin.dispatch` -- all calls before the last one are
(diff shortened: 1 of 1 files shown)


hhugo:
We can release ocamlbuild after this PR is merged and close #322

hhugo:
This is ready for review

gasche:
My understanding is that you ( @hhugo ) have integrated most of the patches from the fdopen repository, and that we believe that the remaining one are not necessary anymore for ocamlbuild to work correctly on Windows. Thanks! Splitting them up, discussing them together and cleaning them up when possible was the right move, I think, and I'm glad I could help from the sidelines.

Do you have an idea of which of the several Windows configurations are known to work correctly for ocamlbuild now? (Do we expect ocamlbuild to work well with or without cygwin? does msvc-or-mingw play any role?)

hhugo:
> Do you have an idea of which of the several Windows configurations are known to work correctly for ocamlbuild now? (Do we expect ocamlbuild to work well with or without cygwin? does msvc-or-mingw play any role?)

What work is what we have in the CI. That is native windows with mingw. I would expect the msys2 setup to work as well (based on the [comment](https://github.com/ocaml/ocamlbuild/pull/339#discussion_r1647804070 from @dra27)

For the cygwin variant, my understanding is that one must use the fdopen repo which is full of patches making it hard to reason about things. What does that mean to be working in that setup ? What set of opam-repo to use ?
I could try to bring back some CI jobs for this setup. Is it still possible with setup-ocaml.3 ?

For the msvc port, I've no idea what to expect. @dra27 can you help here ? is opam supposed to work out of the box in that setup ?

gasche:
Thanks for the clarifications!

Note that I am not myself asking for any particular configuration to work, I was just wondering what the current status is.

hhugo:
#347 demonstrate that msvc is working.
So we have:
- mingw ok
- msvc ok
- cygwin untested, setup-ocaml doesn't support this setup.

PR #341, opened 2024-06-22
Upgrade setup-ocaml

hhugo:


gasche:
"beta" sounds better than "alpha", merged!

PR #344, opened 2024-06-24
Add a tests for long commands 

hhugo:
Add a test that we can handle commands bigger that 8k characters.
Fix the windows implementation that previously had a 8K limit.

### DIFF
--- a/src/my_std.ml
+++ b/src/my_std.ml
@@ -342,6 +342,14 @@ let windows_shell = lazy begin
   shell
 end
 
+let string_exists p s =
+  let n = String.length s in
+  let rec loop i =
+    if i = n then false
+    else if p (String.get s i) then true
+    else loop (succ i) in
+  loop 0
+
 let prepare_command_for_windows cmd =
   (* The best way to prevent bash from switching to its windows-style
      quote-handling is to prepend an empty string before the command name.
@@ -352,7 +360,7 @@ let prepare_command_for_windows cmd =
   (* [maybe_quote] was copied from ocaml/otherlibs/unix/unix_win32.ml *)
   let maybe_quote f =
     if f = ""
-    || String.exists (function ' ' | '\"'| '\t' -> true | _ -> false) f
+    || string_exists (function ' ' | '\"'| '\t' -> true | _ -> false) f
     then Filename.quote f
     else f
   in
--- a/src/my_std.ml
+++ b/src/my_std.ml
@@ -357,20 +357,11 @@ let prepare_command_for_windows cmd =
   let cmd = " " ^ cmd in
   let shell = Lazy.force windows_shell in
   let all = Array.append shell [|"-c"; cmd|] in
-  (* [maybe_quote] was copied from ocaml/otherlibs/unix/unix_win32.ml *)
-  let maybe_quote f =
-    if f = ""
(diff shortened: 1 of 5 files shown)


hhugo:
On windows, the limit seems to be ~8185 which seems very close to the 8192 character limit imposed by CMD.EXE. Yet, we're not longer supposed to go through cmd.exe.
@dra27, any idea where the current limit could be coming from ?

hhugo:
With this PR, I believe we now have all important piece of code of fdopen's patches. The maintenance of ocamlbuild can return to its previous state :)

hhugo:
> With this PR, I believe we now have all important piece of code of fdopen's patches. The maintenance of ocamlbuild can return to its previous state :)

To be more explicit, there are still things present in fdopen patches and absent from upstream but I don't think they are necessary.
- normalize path with forward slash
- set stdout to binary mode to avoid CRLF

src/my_unix.ml
```diff
@@ -58,17 +58,18 @@ let at_exit_once callback =
...
+      Unix.open_process_args_in args.(0) args, cleanup
+    else Unix.open_process_in s, None in
   let close () =
     match Unix.close_process_in ic with
-    | Unix.WEXITED 0 -> ()
+    | Unix.WEXITED 0 -> Option.iter (fun f -> f ()) cleanup
     | Unix.WEXITED _ | Unix.WSIGNALED _ | Unix.WSTOPPED _ ->
-        failwith (Printf.sprintf "Error while running: %s" s) in
+      Option.iter (fun f -> f ()) cleanup;
+      failwith (Printf.sprintf "Error while running: %s" s) in
```
gasche, line 72: I wonder if we would benefit from splitting the two paths more.

Untested:
```ocaml
let close ic = ... in
if not Sys.win32 then begin
  let ic = Unix.open_process_in s in
  Fun.protect ~finally:(fun () -> close ic) @@ fun () ->
  kont ic
else
  let args, cleanup = My_std.prepare_command_for_windows s in
  let ic = Unix.open_process_args_in args.(0) args in
  Fun.protect ~finally:cleanup @@ fun () ->
  Fun.protect ~finally:(fun () -> close ic) @@ fun () ->
  kont ic
```

gasche: (Or whatever style other than `@@ fun () ->` that you prefer or find more consistent with the codebase.)

hhugo: done

src/my_std.ml
```diff
@@ -342,15 +342,74 @@ let windows_shell = lazy begin
...
-  Array.append (Lazy.force windows_shell) [|"-c"; cmd|]
+  let shell = Lazy.force windows_shell in
+  let all = Array.append shell [|"-c"; cmd|] in
+  (* [maybe_quote] was copied from ocaml/otherlibs/unix/unix_win32.ml *)
+  let maybe_quote f =
+    if f = ""
+    || string_exists (function ' ' | '\"'| '\t' -> true | _ -> false) f
+    then Filename.quote f
+    else f
+  in
```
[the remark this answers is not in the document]
hhugo: done

src/my_unix.ml
hhugo: This patch makes the CI complain because it changes the exception raise:
- previously: `Failure ...`
- with this patch: `Finally_raised ..`

I'll revert that part

PR #347, opened 2024-06-25
Add MSVC testing

dra27:
Some work needed on the testsuite, but starting to build on the truly astounding work already done on Windows support here!

Should fix #323 as we're now testing msvc

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Default behaviour, for if core.autocrlf isn't set
+* text=auto
+
+# Quoted strings with newlines
+testsuite/internal.ml text eol=lf
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,11 @@ jobs:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
           allow-prerelease-opam: true
 
+      - name: set ocamlversion
+        run: echo "ocamlversion=$(opam exec -- ocamlc -version)" >> $GITHUB_OUTPUT
+        id: conf
+        shell: bash
+
       - run: opam install . --with-test --deps-only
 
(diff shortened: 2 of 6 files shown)


hhugo:
I've pushed changes to the testsuite. CI is now happy

hhugo:
One still need to properly fix conditional steps in the CI

gasche:
The PR as-is looks fairly clean (unlike some intermediate iterations); congratulations are due to @dra27 and @hhugo. Isn't this ready for review?

hhugo:
No, I'm unhappy about the conditional step change in the ci

hhugo:
@gasche, this is now ready.

dra27:
Thanks, @hhugo - you got the fixes in for the testsuite before I'd had a chance to finish my version locally! Technically [c664696](https://github.com/ocaml/ocamlbuild/pull/347/commits/c6646961e8f0a34a914af80ce07c1b9a54cfb6a0) is not part of restoring MSVC - I don't know if you'd prefer that to be a in a separate PR?

gasche (approved):
The mix of `"foo" -.- bar` and `"foo." ^ bar` in the testsuite is a bit jarring (it took me a few dozen seconds to convince myself than no, they were strictly equivalent), but maybe we can tolerate this if you find it cumbersome to cleanup, and generally the patch is nice to read. I am okay with having the commit that adds .exe to the install script as part of this PR.

hhugo:
> The mix of `"foo" -.- bar` and `"foo." ^ bar` in the testsuite is a bit jarring (it took me a few dozen seconds to convince myself than no, they were strictly equivalent), but maybe we can tolerate this if you find it cumbersome to cleanup, and generally the patch is nice to read. I am okay with having the commit that adds .exe to the install script as part of this PR.

fixed

PR #348, opened 2024-06-25
tests rev-deps

hhugo:
- [x] linux ocaml 5.2 
- [x] linux ocaml 4.14
- [x] windows ocaml 5.2 
- [x] windows ocaml 4.14

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 
 # Quoted strings with newlines
 testsuite/internal.ml text eol=lf
+install-rev-deps.sh text eol=lf
--- a/.github/workflows/rev-deps.yaml
+++ b/.github/workflows/rev-deps.yaml
@@ -1,0 +1,39 @@
+name: rev-deps
+
+on:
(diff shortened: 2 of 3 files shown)


hhugo:
About the windows rev-deps error, they all seem unrelated to ocamlbuild. Here is what I've done so far:
- https://github.com/ocaml/opam-repository/pull/26144
- https://github.com/maranget/hevea/pull/75

hhugo:
@gasche, this is noisy, I don't know if we want to include such test in the CI. We could add it and accept failure, but we probably want to reduce the noise by updating availability on windows inside opam-repository

hhugo:
> @gasche, this is noisy, I don't know if we want to include such test in the CI. We could add it and accept failure of the job, but we probably want to reduce the noise by updating availability on windows inside opam-repository

hhugo:
@gasche, I've looked at the logs, I couldn't find anything that looked like an ocamlbuild issue. I think we can make a release after #349  is merged.

PR #349, opened 2024-06-26
Add macos to the CI

hhugo:


gasche (commented):
I don't understand the relation between the `+unix` changes on OCaml 4 and the macos testing.

hhugo:
> I don't understand the relation between the `+unix` changes on OCaml 4 and the macos testing.

ld complain/warn about inexistant directory in search path
`ld: warning: search path '/Users/runner/work/ocamlbuild/ocamlbuild/_opam/lib/ocaml/unix' not found`

https://github.com/ocaml/ocamlbuild/actions/runs/9678830233/job/26703462729

gasche:
@kit-ty-kate you probably encountered this issue when you dealt with the new-install-directories transition. (Apparently in the code we decided to unconditionally use the new layout.) Could you by chance confirm that the conditional logic proposed by @hhugo is the best approach?

hhugo:
@gasche, the layout change was made by myself in https://github.com/ocaml/ocamlbuild/pull/328

gasche:
Oh, I hadn't realized this and maybe we should have been more careful in #328, but the combined diff makes sense.

src/plugin.ml
```diff
@@ -150,9 +150,14 @@ module Make(U:sig end) =
              about potential regressions introduced by this option.
           *)
 
+          let unix_dir =
+            match Sys.ocaml_version.[0] with
+            | '0' .. '4' -> None
+            | _ -> Some "+unix" in
```
gasche, line 156: See `My_std.split_ocaml_version`.

src/ocaml_specific.ml
```diff
@@ -811,9 +811,17 @@ camlp4_flags' ["camlp4orr", S[A"camlp4of"; A"-parser"; A"reloaded"];
 
 flag ["ocaml"; "pp"; "camlp4:no_quot"] (A"-no_quot");;
 
-ocaml_lib ~extern:true ~dir:"+dynlink" "dynlink";;
-ocaml_lib ~extern:true ~dir:"+unix" "unix";;
-ocaml_lib ~extern:true ~dir:"+str" "str";;
+begin match Sys.ocaml_version.[0] with
+  | '0' .. '4' ->
```
gasche, line 815: (same here)

src/plugin.ml
hhugo: done

src/ocaml_specific.ml
hhugo: done

gasche (approved):

PR #352, opened 2024-07-02
Update ocaml/setup-ocaml to v3

smorimoto:


gasche:
Thanks! Why is there a dependabot file, can we do without it?

smorimoto:
It's pretty annoying to keep manually bumping dependencies, and the goal is to automate with it. I could split it into different PR, but I was just lazy.

gasche:
Forgive my ignorance, I am not used to dependabot. Which dependencies are we talking about? Maybe the dependencies of the github workflow definitions? (which are what? `setup-ocaml`?)

smorimoto:
Yes, Dependabot takes care of version control for GitHub Actions in this case. It automatically opens a PR whenever a new version is released.

smorimoto:
Real world example: https://github.com/ocsigen/tyxml/pull/338

kit-ty-kate:
Thanks!

PR #353, opened 2024-11-27
Get the compiler configuration at runtime (useful for cross-compiling)

### LINKED ISSUE #181: ocamlbuild should not hardcode file extensions
... but rather fetch them from `ocamlc -config`, with the appropriate toolchain.

pirbo:
Fixes #181 (in a minimal way)

The revival of cross compilation by ocaml/ocaml#13526 begs for it ;) 

More seriously, a reproduction scenario cannot really be very minimal as it suppose to have a cross compiler installed but 
1/ suppose you have one thanks to https://github.com/ocaml-cross/opam-cross-windows
2/ do `opam source uutf` and `cd [it]`
3/ try to cross compile it with `ocaml pkg/pkg.ml build --toolchain windows --pkg-name uutf --dev-pkg false` 
(4/ it calls internally 
```
'ocamlbuild' '-use-ocamlfind' '-classic-display' '-toolchain' 'windows'
     '-j' '4' '-tag' 'debug' '-build-dir' '_build' 'opam' 'pkg/META'
     'CHANGES.md' 'LICENSE.md' 'README.md' 'src/uutf.lib' 'src/uutf.cmxs'
     'src/uutf.cmxa' 'src/uutf.cma' 'src/uutf.cmx' 'src/uutf.cmi'
     'src/uutf.mli' 'test/utftrip.native'
```
)
witch fails with
```
Solver failed:
  Ocamlbuild knows of no rules that apply to a target named src/uutf.lib. This can happen if you ask Ocamlbuild to build a target with the wrong extension (e.g. .opt instead of .native) or if the source files live in directories that have not been specified as include directories.
```
because "host" (linux/macos) ocamlbuild hardwired that `ext_lib = .a` while no actually the compiler ocamlbuild would use has `ext_lib = .lib` and therefore it knows how to build `.lib`!

~Ask for feedback:
As it, the test `NativeMliCmi` is broken. This test sets ocamlc to `toto` in order to ensure that any call to ocamlc fails (aka ocamlc is not called, ocamlopt is) and obviously with my patch, ocamlc is called! What should I do? Chance the test? (how?) or make my code way more complex by handling the case where `$OCAMLC -config` fails but `$OCAMLOPT -config` succeed?~ (Thank you @kit-ty-kate )

### DIFF
--- a/src/options.ml
+++ b/src/options.ml
@@ -400,7 +400,10 @@ let init () =
 
   ignore_list := List.map String.capitalize_ascii !ignore_list;
 
-  let raw_ocamlc_config = Command.run_and_read (S [!ocamlc ; A "-config" ]) in
+  let raw_ocamlc_config =
+    try Command.run_and_read (S [!ocamlc; A "-config"])
+    with Failure _ -> Command.run_and_read (S [!ocamlopt; A "-config"])
+  in
   let ocamlc_config_lines = String.split_on_char '\n' raw_ocamlc_config in
   let ocamlc_configs =
     List.filter_map
--- a/src/options.ml
+++ b/src/options.ml
@@ -398,7 +398,27 @@ let init () =
   dir_reorder my_include_dirs include_dirs;
   dir_reorder my_exclude_dirs exclude_dirs;
 
-  ignore_list := List.map String.capitalize_ascii !ignore_list
+  ignore_list := List.map String.capitalize_ascii !ignore_list;
+
+  let raw_ocamlc_config = Command.run_and_read (S [!ocamlc ; A "-config" ]) in
+  let ocamlc_config_lines = String.split_on_char '\n' raw_ocamlc_config in
+  let ocamlc_configs =
+    List.filter_map
+      (fun config ->
+         (* As [raw_ocamlc_config] ends by a ['\n'], [split_on_char]
+            puts an empty string at the end of the list, that's the
+            only line expected to not follow the pattern. *)
+         match String.split_on_char ':' config with
+         | [ k; v ] -> Some (k, String.trim v)
+         | _ -> None)
+      ocamlc_config_lines in
+  let get_ext x =
+    let s = List.assoc x ocamlc_configs in
+    if String.length s > 0 then String.after s 1 else s in
+  ext_lib := get_ext "ext_lib";
+  ext_obj := get_ext "ext_obj";
+  ext_dll := get_ext "ext_dll";
+  exe := List.assoc "ext_exe" ocamlc_configs;
 ;;
 
 (* The current heuristic: we know we are in an ocamlbuild project if
--- a/src/options.ml
+++ b/src/options.ml
@@ -415,13 +415,21 @@ let init () =
          | [ k; v ] -> Some (k, String.trim v)
          | _ -> None)
       ocamlc_config_lines in
-  let get_ext x =
-    let s = List.assoc x ocamlc_configs in
-    if String.length s > 0 then String.after s 1 else s in
-  ext_lib := get_ext "ext_lib";
-  ext_obj := get_ext "ext_obj";
-  ext_dll := get_ext "ext_dll";
-  exe := List.assoc "ext_exe" ocamlc_configs;
+  let get_field k =
+    match List.assoc_opt (k : string) ocamlc_configs with
+    | Some s -> s
+    | None -> failwith (k^" could not be found in ocamlc -config")
+  in
+  let extract_ext s =
+    if String.length s > 0 && s.[0] = '.' then
+      String.after s 1
+    else
+      s
+  in
+  ext_lib := extract_ext (get_field "ext_lib");
+  ext_obj := extract_ext (get_field "ext_obj");
+  ext_dll := extract_ext (get_field "ext_dll");
+  exe := get_field "ext_exe";
 ;;
 
 (* The current heuristic: we know we are in an ocamlbuild project if
--- a/Changes
+++ b/Changes
@@ -14,6 +14,11 @@ of the usual "- ".
 NEXT_RELEASE:
 -------------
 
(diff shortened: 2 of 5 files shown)


src/options.ml
```diff
@@ -398,7 +398,24 @@ let init () =
...
+  ignore_list := List.map String.capitalize_ascii !ignore_list;
+
+  let raw_ocamlc_config = Command.run_and_read (S [!ocamlc ; A "-config" ]) in
+  let ocamlc_config_lines = String.split_on_char '\n' raw_ocamlc_config in
+  let ocamlc_configs =
+    List.concat_map
+      (fun config ->
+         match String.index_opt config ':' with
+         | Some sep -> [String.before config sep, String.after config (sep + 2)]
+         | None -> [])
```
kit-ty-kate, line 410: it's a micro-optimization but concat_map seemed overkill compared to a simple filter_map
```suggestion
    List.filter_map
      (fun config ->
         match String.index_opt config ':' with
         | Some sep -> Some (String.before config sep, String.after config (sep + 2))
         | None -> None)
```

kit-ty-kate:
lgtm but i'm not the most qualified to review this so i'll wait for someone else to review it before merging it.

For the failing `NativeMliCmi` i've used `ocamlopt -config` as a fallback. It's not the best solution because `ocamlc -config` will display `/bin/sh: line 1: toto: command not found` on stderr (because `Unix.open_process_in` is used) but this case should rarely arise if ever so i think this is probably ok to leave as-is for now. If it becomes a problem in the future it can always be fixed.

I'm not sure why the Windows CI fails though, if someone could have a look

src/options.ml
```diff
@@ -412,13 +412,22 @@ let init () =
...
+    if String.length s > 0 && s.[0] = '.' then
+      String.after s 1
+    else
+      s
+  in
   ext_lib := get_ext "ext_lib";
   ext_obj := get_ext "ext_obj";
   ext_dll := get_ext "ext_dll";
-  exe := List.assoc "ext_exe" ocamlc_configs;
+  exe := get_field "ext_exe";
```
pirbo, line 430: Thank you very much for taking over!

That line was not a mistake though (actually I wrote what you wrote at first too and got caught up by the test suite under windows). `exe` is supposed to keep its initial dot while others don't! (Yes I refrained myself to uninformize that as I tried to keep the change minimal but do what you want, you're a maintainer :-) )

kit-ty-kate: it does keep its initial dot though. `get_field` and `get_ext` are two distinct functions

kit-ty-kate: btw did the testsuite work under windows for you? The CI also failed with your commit on Windows

pirbo: My bad!
(I retry on windows right now)

pirbo:
> I'm not sure why the Windows CI fails though, if someone could have a look

It was a trailing `\r` issue...

hhugo (approved):
LGTM

hhugo:
Maybe add a change log entry.

kit-ty-kate (changes requested):
I disagree with the last commit. Renaming `exe` is unnecessary and quite likely breaking users' code. Also the dynamic setting of it was removed. I don't this this is what you want either. I'll revert this change

src/options.ml
kit-ty-kate: i've renamed the two functions and gave them different roles to avoid similar confusion in the future.

kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

kit-ty-kate:
If nobody goes against in the meantime, i'll do a release next week.

pirbo:
Thank you!

I won't fight for it as it "works for me"© :-) (I think the problematic case would be if someone wanted to cross compile **from** Windows) but removing the dynamic setting was my purpose.

Here was my reasoning:
- uutf which I used to test my patch has no myocamlbuild.ml so I tried with ptime which has one plugin.
- It appears that with this patch, ocamlbuild in cross-compilation for windows mode generates and execute on "host" `myocamlbuild.exe` instead of simply `myocamlbuild` (Again, it doesn't matter, linux(/macos) is perfectly happy to run an executable called something.exe, everything works in my case)
- But still, once I realized, intellectual honesty ask me to acknowledge that in `src/plugin.ml` one should not use `Options.exe` which is now the "target" `exe` but an `Option.host_exe` (and as "`Options.host_ocamlc`" is called `Options.plugin_ocamlc` (ocamlc without the `--toolchain ....`) I followed the naming scheme) (If we wanted to be super clean, we could dynamically set it by calling `Options.plugin_ocamlc -config` but I think it's overkill as it cannot change since compilation time.)
- After this change, Options.exe was dead value in the codebase so again dilemma: removing it and introducing a breaking change or keeping it and having an unused therefore untested dangling value...

gasche:
The revdeps testing for the opam 0.16.0 release caught a regression with this PR: the addition of `Command.run_and_read` with a different type than the pre-existing `My_unix.run_and_read` breaks tgsl, which used `run_and_read` in the context `open Ocamlbuild_plugin;; open Command`. See https://github.com/ocaml/opam-repository/pull/27442

We could fix plugin user code to explicitly use `Ocamlbuild_plugin.run_and_read` instead, but I don't really like the idea of breaking compatibility if we can avoid it, especially for such a glacially-moving project as ocamlbuild -- at least we should guarantee stability.

I propose to rename `Command.run_and_read` into `Command.run_and_read_spec`, to avoid the shadowing and reflect the different type. What do you think?

(I would then cancel out the 0.16.0 release and publish 0.16.1 instead.)

gasche:
I have a proposal with `Command.run_spec_and_read` in https://github.com/ocaml/ocamlbuild/pull/361.

pirbo:
Thank you very much for taking care of fixing the regression. I'm also on the page that it had to be done here!

PR #355, opened 2024-11-28
More test

hhugo:


test/test9/test.sh
```diff
@@ -14,6 +14,11 @@
 set -e
 set -x
 cd `dirname $0`/../..
-ocamlc -I src ocamlbuild_pack.cmo test/test9/testglob.ml -o ./testglob.native
+if [ "$OCB" = "ocamlbuild" ]; then
+    IDIR="-I $($OCB -where | tr -d '\r')";
```
kit-ty-kate, line 18: to prevent issues for local runs in case cwd contains characters that are parsed differently by sh
```suggestion
    IDIR="-I '$($OCB -where | tr -d '\r')'";
```

hhugo:
@kit-ty-kate, the CI doesn't like your last commit.

hhugo:
> @kit-ty-kate, the CI doesn't like your last commit.

I've pushed a fix (hopefully)

test/test9/test.sh
```diff
@@ -14,6 +14,11 @@
 set -e
 set -x
 cd `dirname $0`/../..
-ocamlc -I src ocamlbuild_pack.cmo test/test9/testglob.ml -o ./testglob.native
+if [ "$OCB" = "ocamlbuild" ]; then
+    IDIR=('-I' "$($OCB -where | tr -d '\r')")
+else
+    IDIR=('-I' 'src' '-I' 'plugin-lib')
+fi
+ocamlc "${IDIR[@]}" -I +unix unix.cma ocamlbuildlib.cma test/test9/testglob.ml -o ./testglob.native
```
kit-ty-kate, line 22: Array constructions like these ones do not exist in POSIX shell, only in bash and bash-compatible shells.
To use POSIX shell you can do instead:
```suggestion
    set -- '-I' "$($OCB -where | tr -d '\r')"
else
    set -- '-I' 'src' '-I' 'plugin-lib'
fi
ocamlc "$@" -I +unix unix.cma ocamlbuildlib.cma test/test9/testglob.ml -o ./testglob.native
```

PR #356, opened 2024-11-28
Avoid race in makefile

### LINKED ISSUE #248: Using two temporary directories for the generation of “ocamlbuild_pack.cm*”
The directory “tmp” is created by two recipes in [the make script](https://github.com/ocaml/ocamlbuild/blob/7115a9441ed11a736834305ea17a42ab1f251a4d/Makefile#L151 "Update candidate") which is used so far.
I imagine that this approach can become insufficient in the use case that a software build would be attempted in parallel.

How do you think about the possibility to work with two temporary directories for the generation of the desired file variant “ocamlbuild_pack.cm[ox]”?

hhugo:
fix #248

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,22 +142,22 @@ plugin-lib/ocamlbuildlib.cmxa: src/ocamlbuild_pack.cmx $(EXTRA_CMX)
 
 # The packs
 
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Looks obviously correct. Thanks!

PR #360, opened 2025-02-16
add tgls revdep in CI

gasche:
This adds the `tgls` package as a reverse dependency in our continuous integration, as it caught a regression in https://github.com/ocaml/opam-repository/pull/27442 .

(I don't have a sense of the size of its dependencies, so this might be too slow to build, but I don't expect it.)


gasche:
... of course I made the mistake of adding this test to `.travis.yml`, and then being confused because it did not run in the CI. All CI stuff is now in `.github/workflows`, and we should just as well remove `.travis.yml`.

gasche:
This PR "works" as tgls is now tested, but it breaks on the current master due to the `run_and_read` shadowing regression. #361 contains a fix for this regression, so it is the better branch to iterate on the CI test -- I'll merge both the added revdep and the fix at once. Closing here.

PR #361, opened 2025-02-16
Unshadow run_and_read

gasche:
This is a proposal to fix the compatibility breakage due to the shadowing of `My_unix.run_and_read` by `Command.run_and_read`: the new function is now called `Command.run_spec_and_read`.

(I tried to understand why we care about having this function at all, and discovered another place in the codebase where it could/should be used -- we run `ocamlc -where` at runtime to locate the stdlib. I went ahead and used it there as well.)


gasche:
I decided to accept my own proposal :-) I think this is good enough and will let me release a fixed 0.16 version quickly.

gasche:
(The build of `tgls` appears to add 23s on ubuntu CI machines, this is okay.)

PR #362, opened 2025-03-06
Remove symbolic link from the repo

dra27:
The symbolic link in `examples/07-dependent-projects/libdemo` can create problems on Windows - there's mitigations in opam for them (cf. things like https://github.com/ocaml/opam/pull/5793) and workaround needed in `dune pkg` (cf.  https://github.com/Leonidas-from-XIV/ocamlbuild/commit/80cdf58c365bc56d427febf9c99d2423356c2faf).

Simplest approach to me seems to be to delete it and have the `Makefile` infrastructure put it in place. It doesn't look to me as though these examples were ever intended to have `ocamlbuild` invoked directly (i.e. it looks the aim is that the user invokes `make` to run them).

### DIFF
--- a/examples/07-dependent-projects/.gitignore
+++ b/examples/07-dependent-projects/.gitignore
@@ -1,0 +1,1 @@
+libdemo
--- a/examples/07-dependent-projects/Makefile
+++ b/examples/07-dependent-projects/Makefile
@@ -16,11 +16,15 @@ all: native byte
 
 clean:
 	$(OCB) -clean
+	rm -f libdemo
 
-native:
+libdemo:
+	ln -sf ../04-library/libdemo libdemo
+
+native: libdemo
 	$(OCB) main.native
 
-byte:
+byte: libdemo
 	$(OCB) main.byte
 
 profile:
--- a/examples/07-dependent-projects/libdemo
+++ b/examples/07-dependent-projects/libdemo
@@ -1,1 +1,0 @@
-../04-library/libdemo


hhugo (approved):

gasche:
I suspect that I'm going to regret asking, but why are `opam` and `dune` not able to deal with symlinks in source repositories? (In particular, why would `dune` need a change within the testsuite?)

dra27:
It's not the programs, it's the OS - Windows users don't necessarily have the ability to create symlinks which, depending on the exact `tar` used causes errors when extracting the tarball. Dune only inherits this problem with its package management feature (i.e. where Dune is performing the job of opam and compiling dependencies as part of the build)

dra27:
Incidentally, an alternate method would be to remove the `examples` directory from (future) release tarballs, which would be done by adding:
```gitattributes
examples export-ignore
```
to `.gitattributes` in the root directory (see the [`git-archive`](https://git-scm.com/docs/git-archive) man-page for more info)

gasche:
Ooh, nifty, I didn't know about this, nor that git forges would actually use `git archive`. I have a preference for that approach, because being able to call `ocamlbuild` directly on the examples sounds somewhat useful.

gasche:
It looks like the `-h | --dereference` option can be used to ask `tar` to follow symlinks at archive-creation time. The best thing would be to ask `git archive` to use this.

gasche:
@dra27 I implemented your suggestion in 542d6093d590a21a0dfe6a392add69f56f1f42b0.

dra27:
Excellent, thanks!

kit-ty-kate:
the gitattributes rules added were a bit too generic in my opinion so i pushed a follow up commit: https://github.com/ocaml/ocamlbuild/commit/94b9106267548c68b06884ce9872fb4ed88f1986

PR #365, opened 2025-03-10
Run tests with OCaml 5.3

hhugo:


hhugo:
> @hhugo it looks like this is tackling the same issue as #366 but going in a different direction -- and the CI change is good to have as well. Which approach do you prefer?

I like the #366 direction but I feel the current state makes it easy to miss regressions. It would be better to be able to expect a substring in the output.

We could merge this PR (it's cheap) while working on the #366

gasche:
Wise words. Done!

PR #366, opened 2025-03-10
testsuite: support asking for any (unspecified) non-empty error output

gasche:
The internal ocamlbuild testsuite fails in my 5.3 switch because it performs snapshot testing on the error messages produced by the OCaml compiler, whose formatting changed again. (Ah, those pesky compiler devs trying to improve compiler messages...) 

Updating the testsuite whenever compiler error messages change is getting old, so the present PR introduces support in the testsuite for saying: "I expect this test to fail with non-empty output, but I don't care which output".


hhugo:
Replaced by #368

PR #364, opened 2025-03-10
silence new '+ ocamlfind ocamlc -config' messages with -classic-display

gasche:
fixes #363

PR #368, opened 2025-03-19
Refactor testsuite

hhugo:
This is an alternative to #366 
It includes the fix in #364

- Tests can now filter out part of the output.
- Check the output on success as well. This uncovered one typo in the testuite.
- Always check fs expectation on success. Previously not the case if failing_msg was set.  This uncovered another typo in the testuite.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,6 +40,14 @@ jobs:
             ocaml-compiler: "ocaml-base-compiler.4.14.2,system-msvc"
           - os: windows-latest
             ocaml-compiler: "ocaml-variants.5.2.0+msvc"
+        exclude:
+          # The linker emits warnings with old ocaml version on macos
+          # We could try to fix the warnings by passing addition flags
+          # but the value would be limited compared to the effort.
+          - os: macos-latest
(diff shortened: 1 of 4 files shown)


gasche:
The CI are unhappy because there are some linker warnings that break the new, more careful mode.

Generally I like the idea, and the results speak for themselves. I will let you iterate if you think this is worth it, and I'm happy to let you merge whenever the tests do work on our machines and also the CI machines.

hhugo:
> The CI are unhappy because there are some linker warnings that break the new, more careful mode.
> 
> Generally I like the idea, and the results speak for themselves. I will let you iterate if you think this is worth it, and I'm happy to let you merge whenever the tests do work on our machines and also the CI machines.

The CI is now happy. I've disabled testing on macos+ocaml4.13 and macos+ocaml5.0 because I couldn't find an easy way to mute linker warnings. Fixing this would not provide much value anyway.

gasche:
The downside is that this will make the testsuite a bit more fragile -- OSX tools break with a freaky regularity and we will see further warnings that will require more tweaks. But I like the results, merging.

